PR #1, opened 2022-10-13
Add initial repository bootstrapping

dpasque:
## Changes

This PR establishes the initial, overall bootstrapping for the repo. Namely, it adds the basics of a React app, basic build scripts, and things like linting/formatting.

Notably, it does not add...
- Any meaningful functionality whatsoever - the actual app content is just boilerplate junk.
- Redux - we'll add it when we actually add the app state to the application
- GitHub templates - we'll do it later.

Overall, I've tried to strike the (very difficult!) balance between making standards & approaches match things we do elsewhere in A8c (like in Calypso), and also making things fit best practices for the size and scope of the app. 

## Testing

Clone the repo and checkout this branch. Then, play around!

- Run `yarn` to install dependencies.
- Run `yarn start` to see the development build.
- Run `yarn build` to see a prod build output.
- Run `yarn test` to see how tests run. 
- Play around in code files to see the linting.

Also, please let me know any feedback you have!
- Are you good with the linting/formatting approach and rules?
- Anything missing in the package.json and scripts?
- Any concerns with the big-picture decisions? e.g. typescript, css modules (vs css-in-js)


worldomonation (approved):

john-legg:
Looks awesome, Dan! I can tell a lot of work went into this and I'm excited to see the progress 👏 

**issue (non-blocking):** I did have one error pop up when I tried running it locally:

> `ERROR in src/index.tsx:4:17
> TS1261: Already included file name '/Users/john/Projects/bugomattic/src/components/app.tsx' differs from file name '/Users/john/Projects/bugomattic/src/components/App.tsx' only in casing.`

I'm not sure if there's something that I'm missing on my end, but it seems to be a somewhat common bug according to [StackOverflow](https://stackoverflow.com/questions/51197940/file-name-differs-from-already-included-file-name-only-in-casing-on-relative-p) 😆 I just wanted to mention it in case other people run into it.

Some of the suggestions included changing `App.tsx` to a different filename and then changing it back, which didn't work for me. Others suggested making sure `forceConsistentCasingInFileNames` is set to `false` in `tsconfig.json`, but that didn't seem to work for me either. 

I was able to get it working by modifying line 4 of `index.tsx`:  `import App from './components/app';` to `import App from './components/App';`. Not the best solution, but it worked! Have you run into this?

PR #6, opened 2022-11-17
Add basic monitoring (logging and event tracking) infrastructure

dpasque:

This PR adds the basic interfaces and dependency injection for application "monitoring" -- another name for logging and analytic event tracking.

Similar to the API client, there will be two implementations of this API. A local version for development and testing will dump things into the console. The production implementation will wire into Tracks and Logstash.

This adds the basic local implementation and some fake components to test the logging in action.

#### Testing Instructions

- Start the app locally: `yarn start`
- Open the browser console
- You should see a page load event. (You will actually see two! [This is caused by a local React debugging tool](https://stackoverflow.com/questions/61254372/my-react-component-is-rendering-twice-because-of-strict-mode). In production, it fires only once.)
- Click the test buttons and watch the logging and events fly!

#### Issues

Related to #  
Closes #


john-legg (approved):
Looks good to me! Great work as always, Dan 👍

PR #8, opened 2022-12-08
Add the feature selector control

dpasque:

This adds the main UI control for searching for and selecting features!

There are a lot of code changes (sorry! 🙇 ), but not all are very important. The key changes are...
- Updating the normalized reporting config data model so that there are two way references between the main entities. I.e., products store the ids of their sub-features, just as features store the id of their parent. This allows us to to quickly build a hierarchical representation from the same shared state.
- Add a piece to the `reporting-config` redux state slice for tracking the current search term. I waffled a lot on where to put this! I picked the reporting config slice for now because the search term is technically used to filter the reporting config data structure. If you have a better idea, please let me know! 😄 
- Added a new `common` directory that has re-usable library code and components. 
- Added a bunch of new components to the `issue-details` directory for selecting the new feature. The main parent component is `feature-selector.tsx`. I picked this directory because the main goal of these components is to ultimately set some part of the issue details state. Similarly though, please let me know if you think of a better place for these components. 🙂  
- Since we're doing real styling now, I added some core CSS set up. `variables.css` is essentially the "theming" document that has all the colors for the app. The main `index.css` has some relevant global styles now too. 
- I wanted a way to easily test locally with different testing reporting configs. So I've created a way to dynamically fetch different local reporting configs using an environment variable. Set `REACT_APP_REPORTING_CONFIG_NAME` to a JSON name in the `public/local-reporting-configs` directory. For now, mostly to protect accidental info leaking, we will only version control one `default` testing config -- others can just be added on local machines. 

#### Testing Instructions

Unit testing:
- `yarn test:once` (Note, there will be a non-failing console error. This is due to an import dependency issue that doesn't affect the app running and will be fixed in a future PR.)

Visual testing:
- Checkout this branch
- Get a better testing reporting config from me and add it to `public/local-reporting-configs/old-bugomattic.json`
- `REACT_APP_REPORTING_CONFIG_NAME='old-bugomattic' yarn start`

Accessibility testing:
- Use Mac voiceover to interact with the new control
- Try to navigate and use the new tree using your keyboard only

#### Issues

Related to #  
Closes #


src/api/local-api-client.ts
```diff
@@ -12,7 +12,22 @@ export const localApiClient: ApiClient = {
 			} );
 
 		// Let's keep us honest by adding a little delay here.
-		await wait( 3000 );
-		return localReportingConfigJson as ReportingConfigApiResponse;
+		await wait( 2000 );
```
dpasque, line 15: This file is the main change that supports easily switching between local reporting configs with an ENV variable

src/app/app.tsx
```diff
@@ -1,12 +1,6 @@
 import React, { useCallback, useEffect } from 'react';
```
dpasque, line 1: Can ignore: just removing old debug code

src/debug/fake-flow.tsx
```diff
@@ -1,9 +1,6 @@
-import React, { useCallback, useEffect } from 'react';
```
dpasque, line 1: Can ignore: just removing old debug code

src/index.css
```diff
@@ -1,11 +1,44 @@
+@import './variables.css';
```
dpasque, line 1: This is where we set come general global styles for the app. More to come here in the future!

src/issue-details/feature-selector.tsx
```diff
@@ -0,0 +1,59 @@
+import React, { useCallback } from 'react';
```
dpasque, line 1: This is the main parent component of the new control. All the sub components descend from here.

src/reporting-config/__tests__/reporting-config-slice.test.ts
```diff
@@ -21,6 +21,8 @@ describe( '[reporting-config-slice]', () => {
 					product_id: {
```
dpasque, line 21: Can ignore: just matching new state typings

src/variables.css
```diff
@@ -0,0 +1,18 @@
+:root {
```
dpasque, line 1: This is the global theming css of the app

dpasque:
Okay okay, I'm done pushing refactors, I promise! 😆

john-legg (approved):
@dpasque This is awesome! Great work putting all of this together 👏 I looked over the code (to the best of my ability) and nothing stood out as needing to be changed.

By the way, I tested out the accessibility features - so cool!! I did notice that VoiceOver refers to parent tree items as a `button` and children items as a `text element inside a list box`. I wasn't sure if that was intentional or not, but I just wanted to mention it.

@youbrokesomething You might be better at reviewing the accessibility features of Dan's PR 😅 I know you helped Dan with it too!

youbrokesomething:
@john-legg Awesome that you tried out VoiceOver!

I'll take a peek at this tomorrow (Friday).

dpasque:
Thank you so much for the review @john-legg! 🙇 

> parent tree items as a button and children items as a text element inside a list box

Yes, this is currently by design! Because the parent tree nodes aren't actually selectable, they are just buttons that control other display. The actual features are themselves selectable, so they follow the `listbox` -> `option` pattern. I tried `radios` for the features at first, but it wasn't as good of a keyboard experience.  

That said, that may be the wrong design! 😅  @youbrokesomething  -- let me know if something in the experience feels off and if we should go back to something different for the features 👍

youbrokesomething (approved):
This looks and sounds good to me!! Great job!

PR #11, opened 2022-12-13
Move reducer creation into function

### LINKED ISSUE #10: [Bug]: Import circular dependencies can cause a console error about undefined reducer.
### Quick summary

Running the `reporting-config-slice.test.ts` test logs a console error about the `reportingConfig` reduce being undefined. 

This is actually a weird side effects of a circular dependency state we can end up in.

It doesn't impact the production app in a meaningful way, but suggest we should clean up our module structure.

### Steps to reproduce

`yarn test:once`

### Browser

_No response_

### Other notes

This is a side effect of our current code organization.

The challenge is that the test tries to import something directly from the reporting config slice.

This goes...
reporting-config-slice --> issue-details index --> app index --> store ...

Which tries to register the reducers, but we haven't gotten to the exported line in the original file!

This is a perfect storm of...
- colocating state slices with components
- using indexing everywhere
- the reporting config slice not actually having any components
- re-exporting the typed selector and reducer functions.

Again, this won't practically affect the app because...
- We start from App anyway and go down, so all the imports happen right there
- The pointer to the reporting config reduce does actually get set once all the modules finish importing.

But, this is kind of a circular dependency, and we might want to rework our directory structure
-

dpasque:

In short, fixes a problem with how module importing and resolution was working for the `app` directory.

We were mixing two things which was a bad combo...

- `index.ts` style indexing in all our subdirectories
- "Unwrapped" code -- code statements defined at the top level (not wrapped in functions or class definitions), which executes as soon as the module is first loaded.

This means if anyone imported anything from the `app` index, they were running core redux code at import time. This lead to some weird race conditions where redux thought it was getting undefined reducers. This didn't cause production bugs, just console errors, but still pointed to a troubling cocktail.

We're going to keep using the index convention for now, which means that all code in indexed directories should be thoughtfully wrapped (unless it's just exporting constant static values, e.g.)

#### Testing Instructions

- [x] App should still run (`yarn start`)
- [x] Unit tests shouldn't log console errors

#### Issues

Related to #  
Closes #10

### DIFF
--- a/src/app/store.ts
+++ b/src/app/store.ts
@@ -10,15 +10,17 @@ import { ApiClient } from '../api';
 import { issueDetailsReducer } from '../issue-details';
 import { reportingConfigReducer } from '../reporting-config';
 
-const rootReducer = combineReducers( {
-	reportingConfig: reportingConfigReducer,
-	issueDetails: issueDetailsReducer,
-	completedTasks: completedTasksReducer,
-} );
+function createRootReducer() {
+	return combineReducers( {
+		reportingConfig: reportingConfigReducer,
+		issueDetails: issueDetailsReducer,
+		completedTasks: completedTasksReducer,
+	} );
+}
 
 export function setupStore( apiClient: ApiClient, preloadedState?: PreloadedState< RootState > ) {
 	return configureStore( {
-		reducer: rootReducer,
+		reducer: createRootReducer(),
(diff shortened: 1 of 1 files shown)


john-legg (approved):
Looks good, Dan! I was also able to confirm that the unit tests no longer produce an error.

PR #12, opened 2022-12-14
Update task model to match mockups

dpasque:

Updates the task details model to match the current wireframes.

Those details now include three optional properties...

- Link
- Title
- Details

This should allow us a lot of different types of tasks and displays!

#### Testing Instructions

- [x] Unit tests still pass

#### Issues

Related to #  
Closes #


src/reporting-config/types.ts
```diff
@@ -115,7 +115,8 @@ export type LearnMoreLink = SlackLink | P2Link | GeneralLink;
 export type TaskLink = LearnMoreLink | NewGitHubIssueLink;
 
 export interface TaskDetails {
-	instructions?: string;
```
dpasque, line 118: This is the key model update. Everything else is just references.

john-legg (approved):
Everything looks good and the unit tests pass as well! 

I didn't realize you were basing these properties off the wireframes 😅  If you feel like these should be different, please let me know!

PR #14, opened 2022-12-30
Update Feature Selector Styles and Add Tooltips for Descriptions

dpasque:

Makes a few small updates to the feature selector:
- Updates the styles in the tree -- now hovering and selecting use gray backgrounds, and they select the whole row (without indentation)
- Add the descriptions for each element in the tree as tooltips and `aria-descriptions`

#### Testing Instructions

`yarn start`, then...

- [ ] Click around in the feature selector
- [ ] Hover over items that have a description and see the description as the tooltip!

#### Issues

Related to #  
Closes #


john-legg:
@dpasque Thanks for making all of these changes 😄 I tested the tooltips and `aria-despcription` tags and they work flawlessly. Great idea and nice work!! During my testing, I noticed a few minor things:

**Feature Selector**
- After clicking on a parent item in the tree, it remains the hover highlight color (light gray) even after the mouse has moved away.
- The color of the row when selecting an item (dark gray) might be approaching some contrast/readability issues. I know I didn't provide any guidance on what that color should be, so apologies! I will experiment a bit and get back to you. A temporary solution could be to make the text color white.

**Title/Type Selector**
- When VoiceOver reads over the options for _Type_, it triggers the inline error ("You must pick an issue type") on the second radio button (Feature Request), announces the error, and doesn't continue reading the third option.

dpasque:
@john-legg Yayayay, thank you so much for the review! 😄 🙇 

> After clicking on a parent item in the tree, it remains the hover highlight color (light gray) even after the mouse has moved away.

I was using the same style for focus, but I think that's overkill and makes this case confusing. I removed this so it's just a hover style now. 👍 

> The color of the row when selecting an item (dark gray) might be approaching some contrast/readability issues. I know I didn't provide any guidance on what that color should be, so apologies! I will experiment a bit and get back to you. A temporary solution could be to make the text color white.

There was actually another gray color in the a8c scale between the two, so I just took it down one level. I think it looks a lot nicer, and I think that will make that gray color more usable throughout the app. The contrast now is 9.36.

Let me know if you have any other style ideas or preferences for that row, and I can update that! 👍 

Screen Shot 2023-01-02 at 1 13 41 PM

> When VoiceOver reads over the options for Type, it triggers the inline error ("You must pick an issue type") on the second radio button (Feature Request), announces the error, and doesn't continue reading the third option.

Ah, nice catch! 👏  I have a big branch off of this branch, so I'm going to avoid expanding the scope on this PR, so I've made this an issue I can fix on another PR: #15

john-legg:
@dpasque Wow, that was super fast!

> There was actually another gray color in the a8c scale between the two, so I just took it down one level

That color looks great, actually! Nicely done 👏 

> I'm going to avoid expanding the scope on this PR, so I've made this an issue I can fix on another PR: https://github.com/Automattic/bugomattic/issues/15

Sounds good! I'll go ahead and approve this 👍

john-legg (approved):

PR #16, opened 2023-01-02
Clean up the directory structure

dpasque:

This PR makes some long overdue clean-up changes to the directory structure for the app.

**Note!** Clean-up changes like this where there is lots of renaming and moving of things always leads to MASSIVE and confusing diffs. I highly recommend just looking at the branch code and seeing if you like how things are laid out. There are no functional changes to the app in this change.

The clean-up includes...

- Renaming the `FeatureSelector` to the `FeatureSelectorForm` -- this give parity with the `TitleTypeForm`.
- Giving the `FeatureSelectorForm` its own state slice for managing the search term and the temporary selected feature (before you hit "continue" and save it for real).
- Move the two forms to their directories -- they are their own big pieces of the app with their own styles and state, and they should have their own directories.
- Move any combined selectors (where we compute a memoized selector based off of other selectors) to their own files and directory. Cross referencing slice files (like we had been) was getting sloppy.
- (For the most part), ditch the `index` convention ***

*** This convention has just not played well with this app from the start. We are prioritizing [colocation](https://kentcdodds.com/blog/colocation) for our directory structure. However, the combination of colocation and React/Redux often using exported constants has lead to a module mess. We are often way over-importing and sometimes hit dependency race conditions or circular dependencies. We should just avoid that mess! Instead, our flattened, functional-area directory structure should make it easy to just find and import exactly what you need.

I have opted to still allow `index` files when you have a subdirectory that contains all similar file types. (Like a subdirectory of private helper components.) 

#### Testing Instructions

- [x] Unit tests still pass: yarn test:once
- [x] App still runs: yarn start

#### Issues

Related to #  
Closes #


john-legg (approved):
@dpasque Great job cleaning things up! The organization you did looks nice and makes sense. Also, thanks for linking the colocation article 👍 

I ran the app and tests and everything looks good as well.

PR #18, opened 2023-01-05
Add/initial task components

dpasque:

This adds the initial components for the checklist of tasks.

#### Testing Instructions

There are some default tasks created in the default JSON. 

To see them...
1. Boot up the app locally
2. Pick an issue type and hit continue
3. Select a feature

You should then see some tasks! They will change based on the feature and the issue type.

I recommend...
- `Bug + Feature (Under Feature Group)` for seeing a long task to test wrapping
- `Blocker + Feature (Under Feature Group)` for seeing a bunch of different links together

#### Issues

Related to #


john-legg:
Dan, great work on this! I went through the code and everything looks good 👍 

I have a few things that I noticed about a couple of interactions, but perhaps it would be easier to go over them via chat sometime after you get back. By the way, the icons are 💯

PR #22, opened 2023-01-20
Add pointer cursor to all buttons, radios, and checkboxes

dpasque:

For a better interactive feel, we add `cursor: pointer` to all...
- buttons
- radios
- checkboxes

We also make radio and checkbox labels not take the full width so that there's not an obscenely wide click target!

#### Testing Instructions

Boot the app with `yarn start` and click around! The cursor should be a pointer hand when you expect it to :)

#### Issues

Related to #  
Closes #


dpasque:
Abandoning -- turns out, we shouldn't use the cursor pointer!

PR #28, opened 2023-01-26
Add production monitoring client implementation

dpasque:

Adds the production implementation of the monitoring client, which handles logging and analytics (Tracks events).

This doesn't include fully using that monitoring client throughout the app. That will come in a later PR! For now, we just include a few sample places.

#### Testing Instructions

This in many ways depends on backend changes that may/may not be there.

So for now, the best way to test this is with unit tests! 

`yarn test:once`

#### Issues

Related to #  
Closes #

### DIFF
--- a/src/app/app.tsx
+++ b/src/app/app.tsx
@@ -1,5 +1,6 @@
-import React, { ReactNode } from 'react';
+import React, { ReactNode, useEffect } from 'react';
 import { AppHeader } from '../app-header/app-header';
+import { useMonitoring } from '../monitoring/monitoring-provider';
 import { ReportingConfigLoadingIndicator } from '../reporting-config/reporting-config-loading-indicator';
 import { useReportingConfigLoad } from '../reporting-config/use-reporting-config';
 import { ReportingFlow } from '../reporting-flow/reporting-flow';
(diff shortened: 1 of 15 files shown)


src/app/app.tsx
```diff
@@ -1,12 +1,18 @@
...
 import { ReportingConfigLoadingIndicator } from '../reporting-config/reporting-config-loading-indicator';
 import { useReportingConfigLoad } from '../reporting-config/use-reporting-config';
 import { ReportingFlow } from '../reporting-flow/reporting-flow';
 import styles from './app.module.css';
 
 export function App() {
 	const reportingConfigLoadStatus = useReportingConfigLoad();
+	const monitoringClient = useMonitoring();
+
+	useEffect( () => {
```
dpasque, line 13: Our sample usage of triggering a Tracks event

src/next-steps/next-steps.tsx
```diff
@@ -7,17 +7,36 @@ import { MoreInfo } from './sub-components/more-info';
...
+				parentType: tasks[ taskId ].parentType,
+				parentId: tasks[ taskId ].parentId,
+			};
+		} );
+
+		monitoringClient.logger.debug(
+			'Relevant tasks calculated for issue type and feature. See additional details for sourcing.',
+			{ sourcesInOrder }
+		);
+	}, [ relevantTaskIds, tasks, monitoringClient.logger ] );
```
dpasque, line 38: Our sample use of a `debug` logging call. This lets you see where every task you're seeing came from!

john-legg (approved):
The code looks good, the logging works, and the tests pass, including `production-monitoring-client.test.ts`. Nice work, Dan 🚀

PR #30, opened 2023-02-01
Update css to be more resilient to prod deploy

dpasque:

This makes the CSS rules play nicer with CSS conflicts on our MC production deploy.

#### Testing Instructions

If you boot the app, it should look as it always has! Later, we can test this in a phab diff.

#### Issues

Related to #  
Closes #


john-legg (approved):
Looks good to me! Can't wait to see it in MC 🚀

PR #31, opened 2023-02-02
Add URL history tracking for meaningful state in the app

dpasque:

This adds URL history tracking to the app!

At key places in the app, we update the URL with meaningful state. This allows navigating with the browser back/forward button. It also allows bookmarking and sharing URLs with others -- you can basically see the app load to the state you left it in!

To support some limitations of MC, we use query params for all the URL tracking.

#### Testing Instructions

Checkout this branch -> `yarn` -> `yarn start`

- Try going through the app, then using the back/forward browser buttons to undo/redo what you just did.
- Pause where you are in the app, copy the url, and open it in a new tab. You should find yourself in a similar place!
- For convenience, the header takes you back to the start.
- You can try messing with the URL directly if you want! We should be decently resilient to changes (as in we don't crash). The only thing that isn't handled yet (coming in a future PR) is changing out the selected feature ID directly. We'll be centralizing that validation elsewhere in the app. 

#### Issues

Related to #  
Closes #


john-legg (approved):
This is really smooth, Dan! I went back and tested it some more and I couldn't find any issues. Nice work 👍

PR #33, opened 2023-02-08
Add error handling and improved validation

dpasque:

This PR adds some overdue error handling and validation to the app!

It adds...
- A top level error boundary to catch any errors that bubble up in the app
- Handling for when the reporting config fails to load
- Validation of the reporting config as it is being parsed
- Deeper validation (like checking if a feature ID even exists) of state coming in from URL history

Not done in this PR:
We need to update our logic for handling the currently active step. I will be reworking that in another PR! Just worth noting that you could set a bad URL and end up in a weird UI state. 

##### Validation Philosophy

There are lots of tradeoffs and decisions when you add validation and error handling to a site! I want to lay out the general tenets that I followed for transparency...

- **Validate all user input before saving to the store** -- The app should be able to trust what is in the store.  If a state update comes in any way from user input, validate it in the reducer. The main vector for this right now is URL manipulation. 
- **Trust programmatic state updates** -- in contrast to the above, if any piece of state is directly calculated and set by the code, don't do any fancy validation or fallback. If something breaks here, it should break big and loud, because it means there's a bug in our code! Our global error handler will catch the error and notify us. 
- **Reserve the global error handler for unexpected behavior** -- If something falls in the range of expected user interaction, we should handle it gracefully and not throw. Adding fallbacks for URL handling is a good example of this.
-  **Throw liberally for bad reporting configs** -- consider this the one exception to the above! An incorrectly formatted reporting config JSON document just opens up a whole can of worms we don't want. Let's keep on top of that JSON document aggressively!

#### Testing Instructions

I've gone through and tested a lot of permutations of these cases already! There's also a lot of unit/component/integration testing for these changes added in this PR.

If you'd like to play around, some examples of what you can look at...
- Navigating forward and back through the flow and make sure everything works as it did before.
- Manipulate the query params in the URL to set bad data. You may end up in a weird UI state (see note above about reworking the active step logic), but the app should never crash!
- Manipulate local reporting config you are using. You should see the global error display and see an error logged in the console.

#### Issues

Related to #  
Closes #


john-legg:
This looks great, Dan! I followed your testing instructions pretty closely:

- [x] Tested the back/forward functionality. Everything works as expected, even when it involves a broken task.
- [x] Manipulated the URL, which definitely caused some wacky states, but Bugomattic never crashed! 🦾 
- [x] Broke the local config, which caused the global error to display alongside a console error (as expected).
- [x] Tried to crash Bugomattic in so many ways, but couldn't 😄 

**question (non-blocking):**
When a task appears broken (e.g. `Product > Feature Group > Feature (Under Feature Group)` with type `Feature Request`), I see the related error in the browser console repeated about 4-6 times. After inspecting the details of each, there doesn't seem to be a difference between them. Is this just a result of the broken task breaking different things behind the scenes?

Just a random thing I noticed. Everything else looks awesome! 👍

john-legg (approved):

dpasque:
@john-legg 👋 Thank you so much for the review!

> When a task appears broken (e.g. Product > Feature Group > Feature (Under Feature Group) with type Feature Request), I see the related error in the browser console repeated about 4-6 times. 

This is a fantastic catch, and led to a new idea implemented in `46b8f8a` that I think will be really helpful! 🙇 

So, it's a little confusing because _part_ of the verbosity in logging actually comes from a couple of different places that are expected...
1. When running locally, we are using `React.StrictMode`, which helps catch some bad practices at the time of development. However, it also mounts all components twice to do that checking, so you basically get double of everything! For example, this is also why when running locally you see two of the page load events!
2. If any unexpected error occurs in React, the React framework will log a bunch of errors to the console _even if_ you handle that error in an error boundary. Unsurprisingly, this is really controversial and ticks a bunch of people off! 😆 But the React team really stands by it, so I haven't tried to monkey-patch that or anything.

But!! Some of the excess logs were coming because the log call was actually being called on _every rerender_. So, if anything made the component rerender, we would make tons of log spam!

So, I've added a function to create a logger that has a cache! That way, if a log call is a true duplicate from a rerender, we can skip it, and avoid making spam! 😄 

I've tested this so I'll merge for now, but feel free to play around with the new change if you're curious!

john-legg:
Thanks for the thorough explanation, @dpasque! That makes a ton of sense. I just tried it out and there's definitely less duplicate errors coming through. Nice work 👍

PR #34, opened 2023-02-09
Add remaining Tracks events

dpasque:

Adds the actual triggering of the remaining Tracks events on the client!

See this P2 for the full list of events: pciE2j-1LC-p2

** One unique note worth mentioning: Apart from page load, I've tried to tie every event to a direct user action (like actually clicking on the page). This means that loading a URL with existing state or using browser back/forward will _not_ trigger track events. I think this will get us a much more accurate picture of how users are using the app! Otherwise, we could get a lot of event spam in confusing orders.

#### Testing Instructions

You can see the full list of events here: pciE2j-1LC-p2

I have...
1. Gone through the app flow locally and verified all events in the console
2. Added unit tests for all events

You can verify in the console as well!

#### Issues

Related to #  
Closes #

### DIFF
--- a/src/app/__tests__/app.test.tsx
+++ b/src/app/__tests__/app.test.tsx
@@ -96,4 +96,15 @@ describe( '[app]', () => {
 			}
 		);
 	} );
+
+	test( 'On loading, records the "page_view" event', async () => {
+		const apiClient = createMockApiClient();
+
+		const { monitoringClient } = setup( <App />, apiClient );
+
+		await waitForElementToBeRemoved( () =>
+			screen.queryByRole( 'alert', { name: 'Loading issue reporting configuration' } )
+		);
+		expect( monitoringClient.analytics.recordEvent ).toHaveBeenCalledWith( 'page_view' );
+	} );
 } );
--- a/src/combined-selectors/all-tasks-are-complete.ts
+++ b/src/combined-selectors/all-tasks-are-complete.ts
@@ -1,16 +1,34 @@
 import { createSelector } from '@reduxjs/toolkit';
+import { RootState } from '../app/store';
 import { selectCompletedTasks } from '../next-steps/completed-tasks-slice';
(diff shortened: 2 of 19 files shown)


john-legg:
> I've tried to tie every event to a direct user action (like actually clicking on the page). This means that loading a URL with existing state or using browser back/forward will not trigger track events.

Ah, this is genius! I love this approach and I agree it'll keep the event log from getting muddy and unreadable.

I tested all of the events and they worked great 😄 I think the events you have are really solid and I can't wait to see what we can learn from all this!

**thought (non-blocking):**
Thanks for adding the `feature_select` event 🙏 I was wondering if it might be useful to include what product is being selected (before hitting Continue) in a property. The current state (showing the number of features users select before confirming) is super useful and I'm curious if having the additional data might help in figuring out friction points. For example, if users select 10+ features before landing on _Feature XYZ_, what other features did they think it belonged to and is it an issue with other feature names being too generic, the selected feature having a confusing name, etc.?

This is just a random thought though. I'm sure this would add a bit more noise and I know you've worked hard on keeping a good balance of necessary events!

**thought (non-blocking):**
There's probably nothing that can be done, but `right-click + new tab` on a link (task or more info) doesn't trigger a `link_click` event. I guess I have a general distrust of links opening in a new tab, because that's the first thing I did 😆 Note: using `Cmd + click` to open in a new tab triggered the event.

john-legg (approved):

dpasque:
@john-legg Thanks again for the review! 😄 

I pinged you in slack about the first comment, but re:...

> There's probably nothing that can be done, but right-click + new tab on a link (task or more info) doesn't trigger a link_click event.

Alas, there's not really a way to capture this 😞   (without some buck wild hijinks at least haha 😆 )

dpasque:
I tested the feature/product name addition to the tracks events, so I'm merging for now! 👍

PR #36, opened 2023-02-09
Minor updates to the title field (flip, and add info icon)

dpasque:

Ignore the many files, this is actually a relatively small change! (But reference updates touched a lot of files).

This makes the following updates:
- Flips the title and type components so that type comes first before title.
- Changes the name of the step header, and also all the code files, to match this new hierarchy (titleAndType -> typeTitle)
- Changes the title label back to "Title", but adds an info icon with an extended description.
- Changes the warning svg icon to a triangle to make it more distinct from the info svg icon

#### Testing Instructions

Unit tests should still pass and flow should still work as normal!

Main piece of feedback: what do you think of the title tooltip text? Any suggestions?

#### Issues

Related to #  
Closes #


john-legg:
Nice work, Dan! I tested a few things to make sure the flip didn't cause any unwanted issues, but I didn't find anything major. 
The new warning/error icon works really well and it makes it more distinguishable from the other icons. I'm glad you and Brie suggested it!

**nitpick (non-blocking):**
One very minor thing that I noticed during VoiceOver is that reading over the _Type and Title_ form reads the non-flipped version still: `set issue title and type`.

I definitely think the _Title_ tooltip could be _slightly_ reworded to be more concise. However, I'm terrible at website copy, so I don't have a suggestion on alternatives 😅 I'll try to think of something and get back to you!

john-legg (approved):

dpasque:
> One very minor thing that I noticed during VoiceOver is that reading over the Type and Title form reads the non-flipped version still: set issue title and type.

Ooo, nice catch! I also just went and found even more code naming conventions I missed, oops! These are all fixed now 😁

PR #40, opened 2023-03-07
Misc Usability Testing Fixes

dpasque:

This PR is a grab-bag of the smaller fixes that came out of our usability testing. See pciE2j-1Pp-p2 for context.

Fixes include...
- Changing the github task default title to include the repo name.
- Adding an example search to the feature search placeholder text.
- Adding some error messages to the "Next Steps" section if folks URL smash their way to a bad state, or if there's somehow no tasks configured for a given combination.
- Replace the highlighting style in the feature searching with just a bolder font-weight.
- Fix the extra scrollbar that appears in the breadcrumb (TIL, `overflow-x` actually changes the default behavior of `overflow-y`!)

#### Testing Instructions

- [x] Unit tests pass
- [ ] Feel free to `yarn start` and play around with the changes!

#### Issues

Related to #  
Closes #


src/feature-selector-form/feature-selector-form.tsx
```diff
@@ -94,7 +94,7 @@ export function FeatureSelectorForm( { onContinue }: Props ) {
 			<div className={ styles.searchWrapper }>
 				<DebouncedSearch
 					callback={ handleSearch }
-					placeholder="Search for a feature"
+					placeholder='Search for a feature (e.g. "Site editor") '
```
john-legg: Great choice in placeholder text! 👍 
 
**super nitpick (non-blocking):** I feel like the `S` in `"Site editor"` should be lowercase to match the sentence case of the placeholder text. That was my first impression, at least. I'm not an expert on this kind of stuff though 😅

src/reporting-flow/sub-components/next-steps-step.tsx
```diff
@@ -12,19 +15,40 @@ interface Props {
...
 	const isComplete = tasksExist && allTasksAreComplete;
 
 	let stepContentDisplay: ReactNode;
 	if ( tasksExist ) {
 		stepContentDisplay = <NextSteps />;
+	} else if ( requiredInfoIsMissing ) {
+		stepContentDisplay = (
+			<p className={ styles.badStateMessage } role="alert">
+				Hmm... We seem to be missing some information. This almost always means we could not parse
+				the info in the URL. Restart by clicking on the Bugomattic header at the top of the page.
```
john-legg: **suggestion (non-blocking):** Disclaimer: I'm not an expert on copy haha. I'm wondering if we could update the last sentence to make it clearer what the next steps are: 1) recheck the URL for typos or 2) start over.

> "Hmm, we seem to be missing some information. This almost always means we could not parse the info in the URL. Please check the URL for typos or start over using the Bugomattic header at the top of the page."

...or something like that 😅

src/reporting-flow/sub-components/next-steps-step.tsx
```diff
@@ -12,19 +15,40 @@ interface Props {
...
+			</p>
+		);
+	} else if ( noTasksAreConfigured ) {
+		stepContentDisplay = (
+			<p className={ styles.badStateMessage } role="alert">
+				Hmm... It appears this feature area has no issue reporting configuration. You can try
+				reaching out to the team contacts listed below (if available). You can also ping the
```
john-legg: **suggestion (non-blocking):** Disclaimer: I'm not an expert on copy. I'm wondering if this message could be simplified a bit, so it avoids using the `if available` phrase. Perhaps by having a simple message with a contact list that can be added to. Something like:

> "Hmm, it appears this feature area has no issue reporting configuration. You can try reaching out to the contacts listed below for help:"
> - Bugomattic administrators 

If team contacts are available, then they would be added to the list:
> "Hmm, it appears this feature area has no issue reporting configuration. You can try reaching out to the contacts listed below for help:"
> - Example person
> - Example team
> - Bugomattic administrators 

Not sure if that adds more complexity, but just wanted to throw it out there!

john-legg:
@dpasque this looks good! Lots of great changes 👏  

Everything works as expected and I especially like the new highlighting style for feature searching -- it looks _way_ better. I added a few copy-related nitpicks that are super minor, but I'll go ahead and approve it in case you need to move forward with it 🚀 

I also had a few ideas I wanted to run by you in a Slack thread, so look out for that soon!

john-legg (approved):

src/reporting-flow/sub-components/next-steps-step.tsx
dpasque: Ah, that's a good point!!

So, this actually brings up two ideas for me...
1. This is an error state, so we should log an error here, and tell them we are following up on it. We need to correct these bad configs ASAP!
2. I don't feel _super_ great about trying to intelligently parse the More Info for the best contacts, because there's a bit too much flexibility there and it might get complicated and not always pick the right ones. 🤔 But, there will almost always be _something_ in the More Info, so I think we can just say "contact the team" and defer to the judgment of the user. 

So, I'm thinking...

> Hmm, it appears this feature area has no issue reporting configuration. We've been notified and will correct this soon! In the meantime, you can try contacting the product team directly.

What do you think of that?

src/reporting-flow/sub-components/next-steps-step.tsx
dpasque: Sounds good! I don't think we'll want to guide them to fix the URL though, because that will be a losing battle -- the way the state is serialized can be really hard to reproduce by hand! But I'll be more explicit about starting over...

How about:

> Hmm, we seem to be missing some information. This almost always means we could not parse the info in the URL. Please start over by clicking the Bugomattic header at the top of the page.

john-legg: > I don't think we'll want to guide them to fix the URL though, because that will be a losing battle

Ah, that makes sense to me. I think your suggested copy sounds great!

src/reporting-flow/sub-components/next-steps-step.tsx
john-legg: > This is an error state, so we should log an error here, and tell them we are following up on it. We need to correct these bad configs ASAP!

Great point!

> I don't feel super great about trying to intelligently parse the More Info for the best contacts, because there's a bit too much flexibility there and it might get complicated and not always pick the right ones. 🤔 But, there will almost always be something in the More Info, so I think we can just say "contact the team" and defer to the judgment of the user.

Very true! Yeah, that's probably a good call. Just to clarify, is the More Info panel supposed to be visible during the error state? I didn't see it during my testing. 

> Hmm, it appears this feature area has no issue reporting configuration. We've been notified and will correct this soon! In the meantime, you can try contacting the product team directly.

Perfect! I really like that it mentions that we're notified, so the user doesn't feel completely on their own 😆

dpasque:
Alright @john-legg, everything should now be up to date with your design! 👍  (including adding More Info back in). Let me know what other changes and tweaks I should make! 😄

john-legg:
@dpasque this is awesome! Thanks for making these changes so quickly. I know I threw extra stuff your way today 😬

I finished the accessibility testing and everything works as expected. I especially like how it immediately reads the "missing information" and "no reporting configuration" error messages as soon as it lands on the page 👏 

Two final tweaks I'd like to suggest are:

- Reducing the `max-width` of `.badStateImage` just a bit. I put `400px` and it looked good.
- Adding a margin below the _Start Over_ button on the "missing information" error state. I put `2rem` and that seemed about right.

Other than that, I think this PR is ready. Feel free to merge whenever you want 🚀

john-legg (approved):

dpasque:
> Thanks for making these changes so quickly. I know I threw extra stuff your way today 😬

@john-legg thank YOU for all the great feedback as always -- these look sooooooo much nicer now! I really appreciate all your input, never hold back, please! 😁

PR #42, opened 2023-03-07
Add better toolips

dpasque:

Context: pciE2j-1Pp-p2 -- usability testing changes.

This switches away from native browser tooltips (because they are too slow), to custom, more responsive tooltips!

Now...
- The info icons will show a tooltip centered above the icon that appears immediately on hover.
- The feature tree shows tooltips that follow the mouse and appear after one second.

As a side fix, I added CSS to hide the tooltip icons if the screen doesn't support hovering (e.g. smartphones). It felt like bad practice to have an icon you can't actually hover on!

##### Notes on Library Usage
I debated for a while about which library to use. I wanted something stable, but also really simple. These are just tooltips! I really wanted to avoid adding a ton of code and technical complexity to just make a tooltip appear more quickly. 

In the end, [React Tooltip](https://react-tooltip.com/docs/getting-started) seemed the best option:
- It's relatively lightweight
- It's built on [Floating UI,](https://floating-ui.com/) which is rock solid, and might allows us to easily expand to use Floating UI in the future
- Its API is brilliantly simple. 
- It's backed by Digital Ocean currently, which is nice. 

##### Notes on Accessibility
I want to keep using our paradigm of making tooltip content visible to screen readers primarily through ARIA descriptions on the elements they describe. Unfortunately, the one thing this library doesn't make easy is ARIA hiding the entire tooltip. You can easily ARIA hide the contents, but the tooltip wrapper is still visible in the accessibility tree.

For now, I don't think we should worry about this.  If you're using keyboard navigation and a screen reader, the tooltips never appear anyway! And if someone was using a mouse, hovering, and _then_ trying to do screen reader navigation in that hover state (an _extremely_ unlikely edge case), I figured it was better to just be more verbose and have it read the full tooltip rather than have it read an empty tooltip shell! 

#### Testing Instructions

Big questions: how does the styling feel? We can tweak if needed!

#### Issues

Related to #  
Closes #

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,6 +66,7 @@
 		"react-error-boundary": "^3.1.4",
 		"react-redux": "^8.0.4",
 		"react-scripts": "5.0.1",
+		"react-tooltip": "^5.8.3",
 		"url-join": "^5.0.0",
 		"web-vitals": "^2.1.0"
 	},
--- a/src/common/components/form-error-message.tsx
+++ b/src/common/components/form-error-message.tsx
@@ -8,7 +8,7 @@ interface Props {
 
 export function FormErrorMessage( { children }: Props ) {
 	return (
-		<span aria-live="assertive" className={ styles.errorMessage }>
+		<span role="alert" className={ styles.errorMessage }>
 			<ErrorIcon aria-label="Error:" className={ styles.errorIcon } />
 			{ children }
 		</span>
--- a/src/common/components/limited-text-field.tsx
+++ b/src/common/components/limited-text-field.tsx
@@ -8,6 +8,7 @@ interface Props {
 	onBlur?: FocusEventHandler< HTMLInputElement >;
 	onFocus?: FocusEventHandler< HTMLInputElement >;
 	ariaDescribedBy?: string;
+	id?: string;
 }
 
 export function LimitedTextField( {
@@ -17,6 +18,7 @@ export function LimitedTextField( {
 	onBlur,
 	onFocus,
 	ariaDescribedBy,
+	id,
 }: Props ) {
 	const charactersRemaining = characterLimit - value.length;
 	const isOverLimit = charactersRemaining < 0;
@@ -53,6 +55,7 @@ export function LimitedTextField( {
 				onFocus={ onFocus }
 				onBlur={ onBlur }
 				aria-describedby={ ariaDescribedBy }
+				id={ id }
 			/>
 			<p className={ styles.limitMessage }>{ limitMessage }</p>
 		</div>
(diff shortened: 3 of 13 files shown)


john-legg:
@dpasque I love the new tooltips! The react-tooltip library was an excellent choice 👍 

> As a side fix, I added CSS to hide the tooltip icons if the screen doesn't support hovering (e.g. smartphones).

**suggestion (blocking):**
I actually think it would be beneficial to have the tooltip icons (for type and title) visible on mobile, but have the event change to `click` instead. I don't know if having a different tooltip object for desktop vs mobile adds complexity, but I think it would be good to keep the helpful hints for all users. What do you think about that? The feature tree is harder and I'm leaning toward not having tooltips on mobile for simplicity.

> Big questions: how does the styling feel? We can tweak if needed!

**suggestion (blocking):**
Actually, the styling happens to be spot on from the Calypso tooltips, which makes things easy 😅 I did some quick [UX research](https://www.nngroup.com/articles/tooltip-guidelines/) and it appears that there a few tweaks we could do to make it UX-friendly:

- Enabling an arrow on the tooltip, especially when there are multiple targets around it (e.g. feature selection tree).
- Making sure the tooltip doesn't cover any part of the UI that might be needed (e.g. fields, buttons, text, etc.).
- Keeping the tooltip relatively narrow (not too long horizontally, but multiline is fine).

I experimented a bit and here's what I came up with:
```
	<Tooltip
		anchorSelect={ `#${ buttonId } span` }
		delayShow={ 1000 }
		className={ styles.tooltip }
		content={ description }
		place={ 'right' }
	/> 
```
I mostly just changed the anchor to be the text (span) instead of the entire row, removed the `float` and `noArrow` parameters and changed the placement to the right, so it doesn't overlap tree elements or other text. I did this on `expandable-tree-node`, `feature`, and `type-title-form` and it seems to be working okay. I also changed the max width to `20rem` to keep the tooltip relatively compact. Let me know what you think!

I'm still wrapping up some accessibility testing, but I just wanted to send this to you in case you want to try it out!

dpasque:
@john-legg -- I LOVE how these look now with your new suggestions! 😍 Thank you for the detailed research and feedback! 😄  Just pushed up some changes -- want to see if they fit what you were expecting?

john-legg:
> I LOVE how these look now with your new suggestions! 😍 Thank you for the detailed research and feedback! 😄 Just pushed up some changes -- want to see if they fit what you were expecting?

Woo! No problem 👍 I just finished accessibility testing and everything looks good. Your new changes are perfect, so I think this one is ready 🚀 

Edit: I'd also like to mention that clicking the tooltip icons on the _Type and Title_ screen feels _a lot_ better on desktop now that you've added the `click` functionality for mobile. The tooltip used to pop in and out erratically if you clicked on it (I guess it was interrupting the hover state or something). I like that it's nice and smooth now 😄

john-legg (approved):

dpasque:
> The tooltip used to pop in and out erratically if you clicked on it (I guess it was interrupting the hover state or something). I like that it's nice and smooth now 😄

@john-legg That one was really interesting! So, the actual underlying cause was that the icon was originally part of the `label`. I had thought that might make a nicer click target for the inputs, but it actually just made things waaaaaay clunkier. Taking the icons out of the `label` now feels so much better, I agree! 👏 

Thanks again for all the help, your design input was pure gold here, again! 😁

PR #47, opened 2023-03-23
Add reporting config cache

karenroldan:

This PR adds a cache for the reporting config data in the `production-api-client` module. The cache will store the data for 7 days, and if the data is requested within this period, it will be retrieved from the cache instead of the API.

#### Testing Instructions

*

#### Issues

Related to #  
Closes #

### DIFF
--- a/src/api/__tests__/production-api-client.test.ts
+++ b/src/api/__tests__/production-api-client.test.ts
@@ -7,10 +7,16 @@ import { Server } from 'pretender';
 describe( '[ProductionApiClient]', () => {
 	const fakeNonce = 'abc123';
 	const fakeNonceHeaderName = 'x-fake-nonce';
-	const fakeCacheKey = 'cachedReportingConfigData';
+	const cacheKey = 'cachedReportingConfigData';
+	const cacheExpiry = 'cacheExpiry';
 	const fakeReportingConfig: ReportingConfigApiResponse = {
 		foo: {
 			description: 'bar',
+		},
+	};
+	const fakeReportingConfigCached: ReportingConfigApiResponse = {
+		bar: {
+			description: 'foo',
 		},
 	};
 
@@ -21,7 +27,7 @@ describe( '[ProductionApiClient]', () => {
 	beforeEach( () => {
 		globalThis.nonce = fakeNonce;
 		globalThis.nonceHeaderName = fakeNonceHeaderName;
-		localStorage.removeItem( fakeCacheKey );
+		localStorage.removeItem( cacheKey );
 
 		server = createServer( {
 			environment: 'test',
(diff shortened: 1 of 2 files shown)


src/api/__tests__/production-api-client.test.ts
```diff
@@ -62,6 +70,38 @@ describe( '[ProductionApiClient]', () => {
...
+			expect( reportingConfig ).toEqual( fakeReportingConfig );
+
+			const requests = ( server.pretender as Server & { handledRequests: any[] } ).handledRequests;
+			expect( requests.length ).toEqual( 0 );
+		} );
+
+		test( 'Fetches the reporting config again if cache has expired', async () => {
+			const apiClient = createProductionApiClient();
+
+			localStorage.setItem( fakeCacheKey, JSON.stringify( fakeReportingConfig ) );
```
dpasque: **Suggestion (non-blocking)**: Ooo, what if we made this something that was different from the one returned by the fake server, so we can be sure that we do in fact make the request and are not falling back on the cache?

src/api/__tests__/production-api-client.test.ts
```diff
@@ -2,11 +2,12 @@ import { createProductionApiClient } from '../production-api-client';
 import { createServer, Request, Response } from 'miragejs';
 import { ReportingConfigApiResponse } from '../types';
 import { LogPayload } from '../../monitoring/types';
+import { Server } from 'pretender';
 
 describe( '[ProductionApiClient]', () => {
 	const fakeNonce = 'abc123';
 	const fakeNonceHeaderName = 'x-fake-nonce';
-
+	const fakeCacheKey = 'cachedReportingConfigData';
```
dpasque: **Nit (non-blocking)**: tiny nit, but this is actually the real cache key! So I think we can call this just `cacheKey`. It just made me pause for a second and wonder if we were somehow injecting this!

src/api/production-api-client.ts
```diff
@@ -21,7 +27,13 @@ class ProductionApiClient implements ApiClient, LoggerApiClient {
 		const response = await fetch( request );
 
 		if ( response.ok ) {
-			return response.json();
+			const data = await response.json();
+
+			const cacheExpiryTime = Date.now() + 7 * 24 * 60 * 60 * 1000; // 7 days in milliseconds
```
dpasque: **Question (non-blocking)**: As I've been thinking about this, I think I've had a small change of heart -- what do you think if we start with three days? I think I want to start more cautiously, especially as the start will be when there are the most updates to the reporting config. Does that seem like a good balance to you? And sorry for the back and forth! 😆

dpasque (approved):
This reads and tests just fantastically!! 🎉 It feels so snappy in MC now! 😁 I also really appreciate you keeping the tests up to date, and you did such a nice job on the test design!

I left a few very tiny, optional suggestions. Feel free to do them if you want, and then merge away!

src/api/production-api-client.ts
karenroldan: > what do you think if we start with three days?
 
Good idea! We can update it to a longer period after some time. The fix is in [b1c95fd](https://github.com/Automattic/bugomattic/pull/47/commits/b1c95fdc878457a5a47f6ec75e2691ff01df9b02) and updated it to three days.

src/api/__tests__/production-api-client.test.ts
karenroldan: Good catch! I updated the variable and removed the term ´fake´ in [fc82b02](https://github.com/Automattic/bugomattic/pull/47/commits/fc82b029310534d17d908047e8f135dad061337e).

src/api/__tests__/production-api-client.test.ts
karenroldan: Done! I've updated it in [fc82b02](https://github.com/Automattic/bugomattic/pull/47/commits/fc82b029310534d17d908047e8f135dad061337e). Thanks for checking!

PR #49, opened 2023-03-30
Refactor the app to lay foundation for duplicate searching

dpasque:

First of all, I am so so sorry 🙇 😅 

This obviously could have been a lot of smaller PRs, but my goal was just to get the foundations laid and everyone unblocked as soon as possible! As adding duplicate searching obviously expands the scope a lot, I wanted to make some conceptual and structural tweaks here as well that I think make the app a lot easier to digest for folks just onboarding to it.

**_On that note:_** I really recommend **_not_** reviewing this PR by looking at the diffs. It's going to be way to confusing and not very informative! There are so many files changed, but many of those are just renames, re-orgs, or other small tweaks that then affected a lot of files.

Rather, I recommend you just check out the branch locally and consider it more as a whole! 

Also, as UI goes, I just put a bunch of filler that shows the current state. The real components can be made in the coming PRs!

I'll call out some noteworthy things below. 

#### Reworking app boot and hydrating all the data
Relevant directories:
- `app/`
- `static-data/`
- `url-history/`

Bugomattic is a little bit of a funny beast in how it boots up! We have load data from the site from a couple of places.

We have some one time configuration that comes off the server. I've started calling this "static data" is this PR. This includes...
- The "reporting config", or all the feature areas and their issue-reporting details.
- The list of available repo filters when duplicate searching. This is new as of this PR.

We also track the main state of the app in the URL history. This allows people to use bookmarks, alfred workflows, sharing links, etc... So on initial load, we have to parse the state from the URL and load it as well.

The tricky thing is that the URL history load depends on the static data load, as we often have to validate pieces of that URL state against what is in the static data. And we also don't want to show users the app before we finish parsing and loading the URL history state, because then they might see the app just change on them unexpectedly!

So we now show a loading indicator until everything is set up: both the static data and the state from the URL. All of this is now streamlined (so it happens in the right order) in `app/use-app-data-hydration`.

#### Creating Two App Pages
Relevant directories:
- `active-page/`
- `duplicate-searching-page/`
- `reporting-flow-page/`

The app now has two main pages it can display: one for duplicate searching, and one for the reporting flow.

To keep in line with our current state tracking mechanism and to support the route handling in MC, we're just tracking which page is active in a new piece of Redux state.

In the future, we can look at adding `react-router` to handle these main page loads, but it will require reworking some core MC configuration, and some extra reworking of our state tracking.

#### Redux state for duplicate searching:
Relevant directories:
- `duplicate-results/`
- `duplicate-search/`

I've added all the needed redux state and handlers (and some placeholder filler UI components too) for duplicate searching. I've split the state over two top-level nodes:
- One for tracking all the search parameters
- One for tracking the results of the search requests

This keeps the state nodes small and focused. It also helps with tracking state in the URL, as we do it by top-level node, and we only want to track search parameters, not results, in the URL. 

### Testing Instructions

I've updated all the unit tests to get them to passing! ✅ 

You can pull the branch, `yarn`, and `yarn start` to play around with the app locally. Note the UI will be pretty bare bones, but everything will work together, including URL tracking!

I'd also love any feedback on the current state of the repo after these changes: what parts are confusing to understand? Does anything seem over-engineered? Do you have any suggestions for other ways to introduce these foundational changes?

#### Issues

Related to #  
Closes #


src/active-page/active-page-slice.ts
```diff
@@ -0,0 +1,39 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+import { RootState } from '../app/store';
+import { updateStateFromHistory } from '../url-history/actions';
+import { ActivePage } from './types';
+
+const initialState: ActivePage = 'duplicateSearching' as ActivePage;
```
karenroldan, line 6: **suggestion (non-blocking):**  can we use an object to represent the state, even if it has only one property initially? It might be easier to add additional properties to the state object later!

src/app/app.tsx
```diff
@@ -1,41 +1,73 @@
...
+
+	if ( activePage === 'duplicateSearching' ) {
+		return <DuplicateSearchingPage />;
+	}
+
+	return <ReportingFlowPage />;
+}
+
+function AppLoadingIndicator() {
+	const messages = [
```
karenroldan, line 51: **praise:** love the messages here! It's nice to be playful with the messages while the app is loading.

src/app/app.tsx
```diff
@@ -1,41 +1,73 @@
...
+	// We defer any expected error handling (e.g. request failure, bad configuration) to the child components.
+	// This allows as much of the app to be usable as possible, even if some parts are not working.
+	const ready = useAppDataHydration();
+	const activePage = useAppSelector( selectActivePage );
+
+	if ( ! ready ) {
+		return <AppLoadingIndicator />;
+	}
+
+	if ( activePage === 'duplicateSearching' ) {
```
karenroldan, line 43: **nitpick (non-blocking):** should we use constants for these? Just want to make sure we avoid errors caused by inconsistent values, if any, in the future!

src/app/use-app-data-hydration.ts
```diff
@@ -0,0 +1,32 @@
...
+ */
+export function useAppDataHydration() {
+	const dispatch = useAppDispatch();
+	const [ hydrationIsComplete, setHydrationIsComplete ] = useState( false );
+
+	// The order here matters! We validate some state from the URL against configuration from the API.
+	// So we need to load all the configuration first, then update the state from the URL.
+	useEffect( () => {
+		async function hydrateData() {
+			await dispatch( loadStaticData() );
```
karenroldan, line 19: **thought:** _(motivated by the warning from TS compiler)_  if I understand correctly, ´await´ has no effect here since `dispatch` returns the action object that was dispatched. If we want to perform some operation after the action has been dispatched, should we use the `.then()` method?

karenroldan (approved):
I'm always amazed by the changes that you are doing for Bugomattic! 👏🏻  overall, the changes look good to me! I left some suggestions, but all of them are just minor! I tried to understand Redux as I go through the changes. 

> I'd also love any feedback on the current state of the repo after these changes: what parts are confusing to understand? Does anything seem over-engineered? Do you have any suggestions for other ways to introduce these foundational changes?

The changes seem well-organized and manageable. The parts that are confusing to understand are mostly related to minimal (or lack of) knowledge of redux - such as grasping the concept of store, actions, and especially, reducers. It was challenging to wrap my head around the data flow that Redux follows. 

I couldn't think of a better way to introduce the functional changes. What you did was really great! 🎉  You did the foundational changes so quickly. 

The inclusion of the static data, and the list of available repo filters when duplicate searching is a great addition. I like the idea of streamlining it!

src/active-page/active-page-slice.ts
dpasque: Ah, that's a great idea re: future-proofing, and normally I'd be all for it, but this is actually the one state key I was hoping to keep just a string! The reason why is that it makes it a lot easier to manually update the URL!

Because this effectively controls the site navigation, I could see people setting this query parameter manually a lot. For example, you can navigate to `bugomattic&activePage=reportingFlow` and just jump straight into the reporting flow!

With how we do the state serialization in the URL, if we make it an object, the string unfortunately gets a good bit messier. 😕 

Is that okay with you?

src/app/app.tsx
dpasque: That's a good point, and we definitely can! But I think we should be protected by the fact that we're using TypeScript and the type isn't `string`, but rather a union type of two specific strings `'duplicateSearching' | 'reportingFlow'`. So if you try to set it to a different string, it actually will cause a type error and the app won't build!

We use this pattern in a handful of different places throughout the app (for example, filter values, the issue type, etc.). I think in general, I'm more of a fan of using union types than constants because they end up requiring less code and actually block the site from being built if something is wrong.

Let me know what you think! 😄

src/active-page/active-page-slice.ts
karenroldan: > Because this effectively controls the site navigation, I could see people setting this query parameter manually a lot. For example, you can navigate to bugomattic&activePage=reportingFlow and just jump straight into the reporting flow!

That makes sense then! I'm happy with the current implementation. 👍🏻

src/app/use-app-data-hydration.ts
dpasque: Ah, nice catch! This was a typing bug, my bad. 🤦 

In this case, `dispatch` is in fact actually returning a promise! (I find this to be a really confusing part of redux btw), but basically, we use a modified form of dispatch that accepts both actions, which are synchronous, and "thunks", which are basically like async actions.

When `dispatch` accepts a thunk, it returns a promise. When it accepts an action, it just returns the action right away. 

In this case, it should be accepting a thunk, but I had messed up the typings!

Let me know if you have any questions on that -- it definitely took me a while to wrap my head around thunks vs actions 😬

dpasque: And it should be fixed now 🙂

karenroldan: > When dispatch accepts a thunk, it returns a promise. When it accepts an action, it just returns the action right away. 

Ah, now I understand! I was going back and forth on which I should suggest. Thanks for updating it and for the explanation - helps me understand `dispatch` a lot better!

PR #70, opened 2023-04-14
Add the status filter

dpasque:

Adds the status filter! This one is pretty simple and small 😄 

For design reference, see: pciE2j-1VQ-p2#comment-1724

Oh, and it was annoying me, so I updated our "primary" color to be closer to the mock-up!

#### Testing Instructions

I've tested this one a bunch already...

- [ ] Unit tests pass
- [ ] Axe DevTools pass
- [ ] Screen reader works well

Feel free to make sure the design/feel looks good!

#### Issues

Related to #  
Closes #

### DIFF
--- a/src/duplicate-search/duplicate-search-controls.module.css
+++ b/src/duplicate-search/duplicate-search-controls.module.css
@@ -1,5 +1,9 @@
 @import '../variables.css';
 
+.statusFilterWrapper {
+	width: 16rem;
+}
+
 .statusFilterControl {
 	border: 1px solid var( --color-border-dark );
 	border-radius: 2px;
@@ -7,8 +11,6 @@
 	flex-direction: row;
 	align-items: center;
 	overflow: hidden; /* Hide the border radius on the buttons */
-	max-width: 16rem;
-	width: 100%;
 }
 
 .statusFilterControl:focus-within {
--- a/src/duplicate-search/duplicate-search-controls.module.css
+++ b/src/duplicate-search/duplicate-search-controls.module.css
@@ -1,0 +1,51 @@
+@import '../variables.css';
+
+.statusFilterControl {
+	border: 1px solid var( --color-border-dark );
+	border-radius: 2px;
+	display: flex;
+	flex-direction: row;
+	align-items: center;
+	overflow: hidden; /* Hide the border radius on the buttons */
+	max-width: 16rem;
+	width: 100%;
+}
+
+.statusFilterControl:focus-within {
+	border-color: var( --color-primary );
+}
+
+button.statusFilterOption {
+	flex: 1;
+	padding: 0.375rem;
+	border: 1px solid transparent;
+	text-align: center;
+	color: var( --color-text-primary );
+	background-color: var( --color-white );
+}
+
+button.statusFilterOption:focus-visible {
+	outline: none;
+	border: 1px solid var( --color-primary-extra-dark );
+}
+
+button.statusFilterOption:hover {
+	background-color: var( --color-neutral-extra-light );
+}
+
+button.statusFilterOption:active {
+	background-color: var( --color-neutral-light );
+}
+
+button.statusFilterOption[aria-selected='true'] {
+	color: var( --color-white );
+	background-color: var( --color-primary );
+}
+
+button.statusFilterOption[aria-selected='true']:hover {
+	background-color: var( --color-primary-dark );
+}
+
+button.statusFilterOption[aria-selected='true']:active {
+	background-color: var( --color-primary-extra-dark );
+}
--- a/src/duplicate-search/sub-components/status-filter.tsx
+++ b/src/duplicate-search/sub-components/status-filter.tsx
@@ -22,24 +22,28 @@ export function StatusFilter() {
 	const currentStatusFilterOptions: IssueStatusFilter[] = [ 'all', 'open', 'closed' ];
 
 	return (
-		<div
-			aria-label="Issue status filter"
-			role="listbox"
-			aria-activedescendant={ createId( currentStatusFilter ) }
-			className={ styles.statusFilterControl }
-		>
-			{ currentStatusFilterOptions.map( ( statusFilter ) => (
-				<button
-					id={ createId( statusFilter ) }
-					key={ statusFilter }
-					onClick={ createClickHandler( statusFilter ) }
-					aria-selected={ currentStatusFilter === statusFilter }
-					role="option"
-					className={ styles.statusFilterOption }
-				>
-					{ createDisplayText( statusFilter ) }
-				</button>
-			) ) }
+		// This extra wrapper helps with a weird paint glitch in Chromium browsers that can sometimes
+		// leave a single pixel of whitespace within the control.
+		<div className={ styles.statusFilterWrapper }>
+			<div
+				aria-label="Issue status filter"
+				role="listbox"
+				aria-activedescendant={ createId( currentStatusFilter ) }
+				className={ styles.statusFilterControl }
+			>
+				{ currentStatusFilterOptions.map( ( statusFilter ) => (
+					<button
+						id={ createId( statusFilter ) }
+						key={ statusFilter }
+						onClick={ createClickHandler( statusFilter ) }
+						aria-selected={ currentStatusFilter === statusFilter }
+						role="option"
+						className={ styles.statusFilterOption }
+					>
+						{ createDisplayText( statusFilter ) }
+					</button>
+				) ) }
+			</div>
 		</div>
 	);
 }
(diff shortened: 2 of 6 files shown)


src/duplicate-search/sub-components/status-filter.tsx
karenroldan, line 1: **suggestion (non-blocking):** What do you think about rendering the buttons dynamically? I understand that we will only have a maximum of 3 buttons for the filters, but I think for maintainability and consistency, we can render it dynamically. 

We can also use an object for the status filters too. 

Though, I also like the simplicity of the approach and it's easier to understand, so I don't have strong opinions!

src/duplicate-search/duplicate-search-controls.module.css
karenroldan, line 1: **issue (UX,non-blocking):** 

1. I noticed the entire border changes to blue when you select a filter. It will change to black when you click outside. I checked in both Arc Browser (Chromium) and Chrome. 

Screenshot 2023-04-14 at 15 11 04

2. When you select `Closed` from the filter, the button is filled with blue color, but notice that there is a white space at the end of the button. This is only a tiny nit! 

Screenshot 2023-04-14 at 15 15 03

karenroldan (commented):
Hey, Dan! Great work on the implementation! It's always exciting to see the changes happening at Bugomattic and how everything is coming together. 🎉 

I left a couple of comments and all of them are non-blocking!

src/duplicate-search/sub-components/status-filter.tsx
dpasque: Ooo, that's such a good point! I love the idea of creating it dynamically -- it will make it so easy to modify or extend in the future! 😄 

[Change is here](https://github.com/Automattic/bugomattic/pull/70/commits/867201c96e31651f929617eeb5f46aa0c3567403).

> We can also use an object for the status filters too.

This is a good idea and I see the extendability tradeoffs, but if you're okay with it, I think I prefer to keep it just as a string for now, again mostly because this filter is tracked in the URL, and could be one I see people manipulating directly for bookmarks. Is that okay with you?

src/duplicate-search/duplicate-search-controls.module.css
dpasque: > I noticed the entire border changes to blue when you select a filter. It will change to black when you click outside. I checked in both Arc Browser (Chromium) and Chrome.

Ah yes, this is by design! In his Figma mock-ups, John had the outline changing to blue when focused. Is that still the preferred behavior @john-legg ? I can definitely take that out if needed 👍 

> When you select Closed from the filter, the button is filled with blue color, but notice that there is a white space at the end of the button. This is only a tiny nit!

Oh noes! That looks so gross to me 😆 , but i also can't make this happen locally! 🙀  Which browser were you using when you saw that? And did you do anything specifically (like tab in, etc) before you saw that?

src/duplicate-search/sub-components/status-filter.tsx
karenroldan: > I think I prefer to keep it just as a string for now, again mostly because this filter is tracked in the URL, and could be one I see people manipulating directly for bookmarks. Is that okay with you? 

Yep, totally fine with me! The motivation was mainly because I was thinking of how it will be implemented when switching to rendering the button dynamically. 

Your changes with 
```
const createId = ( statusFilter: IssueStatusFilter ) => `status-filter-${ statusFilter }`;
```
 is more ideal!

karenroldan (approved):
_I forgot to select Approve when I submitted my review!_ Feel free to merge whenever! 🚀

src/duplicate-search/duplicate-search-controls.module.css
karenroldan: > Ah yes, this is by design! In his Figma mock-ups, John had the outline changing to blue when focused. 

Oh, okay! My bad! 

> Oh noes! That looks so gross to me 😆 , but i also can't make this happen locally! 🙀 Which browser were you using when you saw that? And did you do anything specifically (like tab in, etc) before you saw that? 

I checked with Arc Browser and Chrome! It happens when you click on `Closed` button.

dpasque: Lol, I'm so sorry, I accidentally clicked the thumb down emoji lololol 🤦  I am very glad we fixed this in fact! 😆

john-legg: > In his Figma mock-ups, John had the outline changing to blue when focused. Is that still the preferred behavior @john-legg ? 

Yup, that's correct! 👍

src/duplicate-search/duplicate-search-controls.module.css
```diff
@@ -0,0 +1,53 @@
...
+	border: 1px solid var( --color-border-dark );
+	border-radius: 2px;
+	display: flex;
+	flex-direction: row;
+	align-items: center;
+	overflow: hidden; /* Hide the border radius on the buttons */
+}
+
+.statusFilterControl:focus-within {
+	border-color: var( --color-primary );
```
john-legg, line 17: 👋 Hey @dpasque, sorry for chiming in late. 

**nitpick (if-minor):** Super nitpick that might have been a conscious decision, but the border for the focused state actually increases to `1.5px` in the designs. I don't know if `outline` is less desirable, but it seems to add thickness without increasing the size of the component:
```
    border-color: var( --color-primary );
    outline: 0.5px solid var( --color-primary );
```

What do you think? If that seems good, I can make the change if you have a lot on your plate.

john-legg:
Hey @dpasque sorry for the late review. Everything looks great 👍 I just had a few minor things that I noticed:

**nitpick (if-minor):** When using `Tab` to focus on the segments, it looks the border for the focused segment is clipped a bit by the segment next to it:
Screenshot 2023-04-14 at 2 38 09 PM
I can't really figure out why that is, maybe you'll be able to spot it faster 😆 

**issue (non-blocking):** In VoiceOver, I noticed that after selecting an option (e.g. "Closed), it'll read out "Closed, selected." If I go to select a second, different option (e.g. "Open"), it doesn't read it out -- I just hear the selected chime. I don't know if this is a VoiceOver quirk, but I just wanted to mention it! 

Note: To clarify, it seems to happen on the second selection and beyond.

src/duplicate-search/duplicate-search-controls.module.css
dpasque: Ack, sorry I merged too early! I had been mostly focused on the color change, so went ahead after Karens review, my bad!

Oh sure, that's a great idea! 😄  I can spin up a new PR with the fixes!

My main question on this bit: are you okay if we just make it 1px? I'm not a huge fan of half pixels in CSS... they're rendered differently by different browsers and screen densities, and I think it's often simpler to avoid them if possible! Does that throw off the design mojo too much?

john-legg: > Ack, sorry I merged too early! I had been mostly focused on the color change, so went ahead after Karens review, my bad!

No worries!

> My main question on this bit: are you okay if we just make it 1px?

I have no problem with that! I didn't know that about half pixels. I'll make the change in the Figma file as well. Thanks for all of your expertise on this 🙇

dpasque:
Hey John! 😄 Great questions!

>When using Tab to focus on the segments, it looks the border for the focused segment is clipped a bit by the segment next to it

"It's an illusion, Michael!" 😆 This is actually a goofy trick-on-the-brain thing. The border isn't cut off, it's just that it's a similar color to the border for the whole control, so it makes those touching sides seem like a thicker border, if that makes sense.  Really the best way around that would be to pick a different `:focus-visible` kind of style. We could, for example, opt for a different color? Or do an inset box shadow? Any thoughts there?

>In VoiceOver, I noticed that after selecting an option (e.g. "Closed), it'll read out "Closed, selected." If I go to select a second, different option (e.g. "Open"), it doesn't read it out -- I just hear the selected chime. I don't know if this is a VoiceOver quirk, but I just wanted to mention it!

Hmm, I'm having a hard time reproducing this 😅  -- could you share more about the keyboard actions you're using??

src/duplicate-search/duplicate-search-controls.module.css
dpasque: Got a fix PR started here: https://github.com/Automattic/bugomattic/pull/72

john-legg:
> "It's an illusion, Michael!" 😆

LOL that was perfection 👌 I never thought I would see that reference in a GitHub PR 😆 

> The border isn't cut off, it's just that it's a similar color to the border for the whole control, so it makes those touching sides seem like a thicker border

Yeah, that makes way more sense. I see it now! I'll have to experiment and see what works the best, but I think what you have is perfectly fine for now.

> Hmm, I'm having a hard time reproducing this 😅 -- could you share more about the keyboard actions you're using??

Now that this PR is merged into trunk, I'm having a bit of trouble reproducing it as well 😅 mostly because selecting an option goes immediately into duplicate searching. Seems like it's not an issue anymore!

PR #71, opened 2023-04-14
Redesign Next Steps

### LINKED ISSUE #53: [Task - Reporting Flow]: Redesign the Next Steps tasks to be cards with CTAs
### Details

Redesign the Next Steps tasks so that they are cards with CTA buttons (if there are links). They will also continue to be checkmarks.

See: pciE2j-1VQ-p2#comment-1674

### Checklist

- [ ]

karenroldan:

Context: pciE2j-1VQ-p2#comment-1674. 

Links on the title are removed. A CTA button is added for tasks with links instead. 

##### BEFORE
before

##### AFTER
**Task 1: Long title with link. Task 2: Long details without link**

Screenshot 2023-04-14 at 16 42 50

**All tasks with links**
 
Screenshot 2023-04-14 at 14 19 57

#### Testing Instructions

Working on CSS is another level 😅 I tried to fix all the possible scenarios! Tests are already updated according to the changes.

- Make sure that when clicking the button, you are redirected to the right page. 
- The trigger to checkbox change is still there: make sure that the checkbox is also checked when clicking the button.

**axe DevTools**
_This will be my first time using this extension._

- [Test URL](http://localhost:3000/Automattic/bugomattic?activePage=reportingFlow&duplicateSearch%5BsearchTerm%5D=&duplicateSearch%5BstatusFilter%5D=all&duplicateSearch%5Bsort%5D=relevance&issueDetails%5BfeatureId%5D=Product__Feature%20Group__Feature%20%28Under%20Feature%20Group%29&issueDetails%5BissueType%5D=bug&issueDetails%5BissueTitle%5D=&activeReportingStep=nextSteps)
- No automatic issues found 

#### Issues

Related to #  
Closes #53


karenroldan:
I see from Figma that the app icon for Slack is colored and the link icon is blueish. I tried overriding the color using CSS but the changes are not reflecting. 

@dpasque can I ask for guidance on this? Do I need to add new SVG files? 

cc: @john-legg

dpasque:
@karenroldan hi! 😄 

> I tried overriding the color using CSS but the changes are not reflecting.
> @dpasque can I ask for guidance on this? Do I need to add new SVG files?

Yeah, you can't really override the SVG color with css easily, so you just need new SVG content! If you open the file in Figma, in the sidebar you can right click the item for the icon and click "copy as SVG". Although, it looks like the Slack icon might be an image in Figma.

For reference, here's the SVG code for the blue link icon:
```
<svg width="20" height="20" viewBox="0 0 20 20" fill="none" xmlns="http://www.w3.org/2000/svg">
<mask id="mask0_671_16950" style="mask-type:luminance" maskUnits="userSpaceOnUse" x="2" y="2" width="16" height="16">
<path fill-rule="evenodd" clip-rule="evenodd" d="M10.8333 4.16667V2.5H17.5V9.16667H15.8333V5.345L9.33917 11.8392L8.16083 10.6608L14.655 4.16667H10.8333ZM15.8333 15.8333V10.8333H14.1667V15.8333H4.16667V5.83333H9.16667V4.16667H4.16667C3.24583 4.16667 2.5 4.9125 2.5 5.83333V15.8333C2.5 16.7542 3.24583 17.5 4.16667 17.5H14.1667C15.0875 17.5 15.8333 16.7542 15.8333 15.8333Z" fill="white"/>
</mask>
<g mask="url(#mask0_671_16950)">
<rect width="20" height="20" fill="#0675C4"/>
</g>
</svg>
```

And for Slack, we probably should try to use their official one from their media kit!

It looks like there's an SVG (towards the bottom) on their collection here: https://salesforce.widencollective.com/c/xt57vnlf

src/next-steps/next-steps.module.css
```diff
@@ -8,18 +8,46 @@
 
 .task {
 	display: flex;
+	padding-right: 10px;
```
dpasque: **Nit (non-blocking):** For all the padding values here, do you mind updating them to be in `rem` instead of `px`? You can leave all the borders etc as pixels, I've just been trying to do all font-sizes and padding/margins in `rem` so things scale with font scaling a bit easier! 😄

karenroldan:
> Yeah, you can't really override the SVG color with css easily, so you just need new SVG content!

Got it! Thanks, Dan! 

I've added new SVG content for the external link and Slack. Changes can be found here: [cbdeee6](https://github.com/Automattic/bugomattic/pull/71/commits/cbdeee63226a20e235444e058c90a1fc7e26b0a9)

Screenshot 2023-04-14 at 16 35 41

src/next-steps/next-steps.module.css
```diff
@@ -8,18 +8,46 @@
...
+	border: var( --color-primary ) 1px solid;
+	border-radius: 4px;
+	padding: 10px 20px;
+	display: inline-flex;
+	align-items: center;
+	font-size: var( --font-body );
+	color: var( --color-primary );
+	text-decoration: none;
+	font-weight: 500;
 }
```
dpasque, line 51: **Suggestion (should be done before merge, but not urgent):** Per the Figma component guide, it looks like the these kind outline fake-button styles should have a `hover` effect, and I think it would be good to add an `active` one too (it adds a sense of "weight" when clicking).

Our current color scheme is a bit off from the Figma ones, and we'll just go back and reconcile that later, so for now, this should work:

```css
a.taskLink:hover {
	background-color: var( --color-neutral-light );
}

a.taskLink:active {
	background-color: var( --color-neutral );
}
```

src/next-steps/next-steps.module.css
```diff
@@ -8,18 +8,46 @@
...
-	margin-top: 2rem;
+	margin-top: 1rem;
 }
 
 .taskTitle {
 	font-size: var( --font-header );
 	font-weight: 500;
 	color: var( --color-text-primary );
+	word-wrap: break-word;
+	word-break: break-all;
```
dpasque: **Suggestion (should be done before merge, but not critical)**: I think the `word-wrap` is a great idea here, in case there's a really long URL or something like that! But I think we should avoid using the `word-break` if at all possible -- I don't think it will be necessary, and it can make the task a bit harder to read if it's long!

src/next-steps/sub-components/task.tsx
```diff
@@ -76,26 +76,28 @@ export function Task( { taskId }: Props ) {
...
 					href={ href }
 					rel="noreferrer"
 					// When they open a link, let's trigger the checkbox change too
 					onClick={ handleLinkClick }
 				>
 					{ getAppIconForLink( link ) }
-					<span className={ styles.linkText }>{ linkText }</span>
+					<span className={ styles.linkText }>{ getLinkName( link ) }</span>
 					<LinkIcon aria-hidden={ true } className={ styles.linkIcon } />
 				</a>
```
dpasque: **Suggestion (non-blocking)**: Re: accessibility, I think what is here is fine, but having the link only say the link title on a screen reader feels a little terse, and could maybe use a bit of context!

I wonder if using `aria-describedby` and using the issue title might be the best way forward?

Here is what that could look like:

``` typescript
const titleId = removeSpaces(`title-${ taskId }`);
titleDisplay = <span id={titleId} className={ styles.taskTitle }>{ titleText }</span>;
buttonDisplay = (
<a
	className={ styles.taskTitle }
	className={ styles.taskLink }
	target="_blank"
	href={ href }
	rel="noreferrer"
        aria-describedby={titleId}
	// When they open a link, let's trigger the checkbox change too
	onClick={ handleLinkClick }
>
	{ getAppIconForLink( link ) }
	<span className={ styles.linkText }>{ linkText }</span>
	<span className={ styles.linkText }>{ getLinkName( link ) }</span>
	<LinkIcon aria-hidden={ true } className={ styles.linkIcon } />
</a>
```

That way, the screen reader will read: "Link: Github. <The title>" which seems nice to me!

dpasque (approved):
Awesome job on this one Karen!! 🎉  😄 It looks a feels really nice!  I have a few comments, and a few that should be done pre-merge, but once those are done, you should be good to go!

dpasque:
@karenroldan and @john-legg -- it's definitely non-blocking for this PR, but one thing I noticed was that these CTA card/buttons don't really scale down on mobile very well right now. MC doesn't even adapt to mobile lol, so it's not super important, but I was thinking we should probably go back and tweak that eventually -- it would be ideal to keep the site responsive!

@john-legg -- I didn't see any mock-ups re: the mobile design, did you have any ideas there? Just move the button underneath maybe?

src/next-steps/next-steps.module.css
karenroldan: Yes! I realized late how bad it looked with the `word-break`. I've updated the property and removed both `word-wrap` and `word-break` and replaced it with `overflow-wrap: break-word;`. Changes can be found here: [7b2196c](https://github.com/Automattic/bugomattic/pull/71/commits/7b2196cf0cfdf0b82829ac0520dd3aecc47904df)

src/next-steps/next-steps.module.css
karenroldan: Nice catch! I've added the CSS in [85c21db](https://github.com/Automattic/bugomattic/pull/71/commits/85c21db2c753083288de429f922cee7293939356)

src/next-steps/next-steps.module.css
karenroldan: Done! Updated all padding values to use `rem` instead of `px` in [85c21db](https://github.com/Automattic/bugomattic/pull/71/commits/85c21db2c753083288de429f922cee7293939356)

john-legg:
> Per the Figma component guide, it looks like the these kind outline fake-button styles should have a hover effect

> Our current color scheme is a bit off from the Figma ones

@karenroldan @dpasque apologies for confusion over new styles vs old styles. The component guides in the current Figma file are all updated for the "new" style, which Duplicate Search will be implementing. In the development guides, I kept the style of Report Issue the same to make iterations a bit simpler by not throwing in styling 😆 

I plan to update the designs for Report Issue with the new styles, but if you'd like to make those changes (e.g. button styling) now, feel free to do so! If it causes too much confusion, I can also update them early. Let me know what you all prefer!

> I didn't see any mock-ups re: the mobile design, did you have any ideas there? Just move the button underneath maybe?

Yes, that seems to be the best option based on experimentation. Thanks for bringing this up! I didn't even think about mobile for the new cards 😅 It would look something like this:
Screenshot 2023-04-14 at 3 26 54 PM

src/next-steps/sub-components/task.tsx
karenroldan: Thanks for your suggestion! It was interesting to learn how to implement better accessibility! I made my changes here: [adcffec](https://github.com/Automattic/bugomattic/pull/71/commits/adcffeca79cf0e8744a0dd030778d702310c57c5)

karenroldan:
Thanks for the review, @john-legg and @dpasque. I will create another PR to reflect the changes for mobile!

PR #79, opened 2023-04-21
Add the search input, and catch up on some tests

### LINKED ISSUE #63: [Task - Duplicate Searching]:  Put the final page component together (with a search)
### Details

pciE2j-1VQ-p2#comment-1693

### Checklist

- [ ]

dpasque:

- Adds the search input to the duplicate searching page
- Removes much of the placeholder filler
- Makes search inputs match the style guide more
- Adds handling to not search if there's no search term
- Backfills the component tests for search results over the course of the search lifecycle

#### Testing Instructions

- [x] Unit tests pass
- [x] No Axe devtool errors

#### Issues

Related to #  
Closes #63

### DIFF
--- a/src/duplicate-search/duplicate-search-controls.tsx
+++ b/src/duplicate-search/duplicate-search-controls.tsx
@@ -15,7 +15,7 @@ export function DuplicateSearchControls() {
 
 	return (
 		<section>
-			<div>
+			<div className={ styles.searchWrapper }>
 				<DebouncedSearch
 					placeholder="Search"
 					callback={ handleSearchTerm }
--- a/src/duplicate-search/duplicate-search-controls.tsx
+++ b/src/duplicate-search/duplicate-search-controls.tsx
@@ -17,7 +17,7 @@ export function DuplicateSearchControls() {
 		<section>
 			<div className={ styles.searchWrapper }>
 				<DebouncedSearch
-					placeholder="Search"
+					placeholder="Search for duplicate issues"
 					callback={ handleSearchTerm }
 					inputAriaLabel="Search for duplicate issues"
 					debounceMs={ 500 }
--- a/src/duplicate-search/duplicate-search-controls.tsx
+++ b/src/duplicate-search/duplicate-search-controls.tsx
@@ -21,6 +21,7 @@ export function DuplicateSearchControls() {
 					callback={ handleSearchTerm }
 					inputAriaLabel="Search for duplicate issues"
 					debounceMs={ 500 }
+					debounceCharacterMinimum={ 3 }
 				/>
 			</div>
 			<div className={ styles.filterSortBar }>
--- a/src/duplicate-search/duplicate-search-controls.tsx
+++ b/src/duplicate-search/duplicate-search-controls.tsx
@@ -1,35 +1,28 @@
 import React from 'react';
-import { useAppDispatch, useAppSelector } from '../app/hooks';
-import { selectAvailableRepoFilters } from '../static-data/available-repo-filters/available-repo-filters-slice';
-import {
-	setActiveRepoFilters,
-	setSearchTerm,
-	setSort,
-	setStatusFilter,
-	setSearchParam,
-} from './duplicate-search-slice';
 import { RepoFilter, StatusFilter } from './sub-components';
+import { useAppDispatch } from '../app/hooks';
+import { setSearchTerm, setSearchParam } from './duplicate-search-slice';
+import { DebouncedSearch } from '../common/components';
 import styles from './duplicate-search-controls.module.css';
 
 // TODO: This is a placeholder component for the duplicate search controls. Modify and tweak however needed! :)
 export function DuplicateSearchControls() {
 	const dispatch = useAppDispatch();
-	const repoFilters = useAppSelector( selectAvailableRepoFilters );
 
-	const handleFauxSearch = () => {
-		dispatch( setSearchTerm( 'Test search term' ) );
-		dispatch( setStatusFilter( 'open' ) );
-		dispatch( setActiveRepoFilters( [] ) );
-
-		// What the dispatch in most components will actually look like
-		dispatch( setSearchParam( setSort( 'relevance' ) ) );
+	const handleSearchTerm = ( searchTerm: string ) => {
+		dispatch( setSearchParam( setSearchTerm( searchTerm ) ) );
 	};
 
 	return (
 		<section>
-			<p>This is just a filler component</p>
-			<button onClick={ handleFauxSearch }>Faux search</button>
-			<p>Available repo filters: { repoFilters.join( ', ' ) }</p>
+			<div>
+				<DebouncedSearch
+					placeholder="Search"
+					callback={ handleSearchTerm }
+					inputAriaLabel="Search for duplicate issues"
+					debounceMs={ 500 }
+				/>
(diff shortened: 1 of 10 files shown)


src/duplicate-search/duplicate-search-slice.ts
karenroldan, line 1: **suggestion:** To minimize the number of search requests and reduce the amount of returned results, should we implement a minimum character limit before initiating the search, such as 3 characters? This approach can help prevent performance issues, especially when dealing large datasets.

karenroldan (approved):
Great job on the changes! I've tested them locally, and everything works smoothly. Also, the tests you've written cover scenarios that we need to account for! 🎉 

**thought:** While the current implementation prevents a new search when the user clears the input field, it might be beneficial to display the initial search state or clear the current results. In that way, the search results also reflect the current input. What do you think?

src/index.css
```diff
@@ -35,18 +35,24 @@ body {
...
 	background-color: var( --color-white );
-	border: 2px solid var( --color-border-dark );
+	border: 1px solid var( --color-border-dark );
 	box-sizing: border-box;
 }
 
+.root input[type='text']:hover,
+.root input[type='search']:hover,
+.root input[type='email']:hover {
+	outline: 1px solid var( --color-border-dark );
```
john-legg: I feel bad because I actually changed the hover style in Figma on Friday without realizing you were doing a PR for this 😅 So sorry about that! 

I changed it to `background-color: var(--color-neutral-extra-light );`. The border/outline hover style was drawing a bit more attention than I was expecting and this updated style is more consistent with all the other components. Would you be willing to change it even though you just did the work for this one? 🙇

john-legg: Also, in regards to hover states, it looks like the search icon has a hover state with a similar style to a focus state (e.g. `border: 2px solid var( --color-primary )`). Would it be possible to remove that and just use the hover state for search bar itself or is there an accessibility-related reason to have both?

Screenshot 2023-04-24 at 4 42 27 PM

src/duplicate-search/duplicate-search-controls.tsx
```diff
@@ -1,34 +1,28 @@
...
 	};
 
 	return (
 		<section>
-			<p>This is just a filler component</p>
-			<button onClick={ handleFauxSearch }>Faux search</button>
-			<p>Available repo filters: { repoFilters.join( ', ' ) }</p>
+			<div className={ styles.searchWrapper }>
+				<DebouncedSearch
+					placeholder="Search"
```
john-legg: **suggestion (non-blocking):** It'd be nice to have a slightly more descriptive placeholder to [help users understand what they can search for](https://www.smashingmagazine.com/2019/06/better-search-ux-microcopy/#placeholder-text), similar to what you did for the search bar in feature selector. Another reason is because the descriptive "empty state" that users first see won't always be there to guide them after they make their first search. I don't know what that copy would be other than `Search for duplicate issues` lol, but there's probably something 😆

john-legg (approved):
Hey @dpasque 👋 This looks great! I made a couple of minor style-related comments, but that's it. Accessibility testing checks out as well. Nicely done 👍

src/index.css
dpasque: > Would it be possible to remove that and just use the hover state for search bar itself or is there an accessibility-related reason to have both?

Hmm, that's a good question! 🤔 The border on hover definitely looks goofy -- we should change that for sure! There's not really accessibility concerns per se because we actually remove that button from the accessibility tree (it's just for the mouse), but the challenge is that there's not really a good way to apply the hover style to the whole input when you hover over the button (this is a side effect of how you have to absolutely position buttons over text inputs).

As a middle ground, how would you feel if we applied to the search button the basic hover=neutral-extra-light, active=neutral-light that we do to the rest of our other clear buttons?

dpasque: Btw, I've pushed that up if you want to take a look and play around with it!

dpasque:
@john-legg @karenroldan hi! 👋 😄 

Karen, I love the two ideas you brought up above! 😄 

> While the current implementation prevents a new search when the user clears the input field, it might be beneficial to display the initial search state or clear the current results. In that way, the search results also reflect the current input. What do you think?

> To minimize the number of search requests and reduce the amount of returned results, should we implement a minimum character limit before initiating the search, such as 3 characters? This approach can help prevent performance issues, especially when dealing large datasets.

John, I was wondering if you had any strong opinions or other design guidance for these two ideas?

Re: the minimum character limit, I've done things in the past where the debounce (searching after a delay) is restricted to a character limit, but you can always press enter (or the search button) to override that and fire the search anyway. I wonder if that might be a good middle ground?

src/index.css
john-legg: > Hmm, that's a good question! 🤔 The border on hover definitely looks goofy -- we should change that for sure!

Gahhh, you're going to _hate_ me, but even though I tested `--color-neutral-extra-light` as the hover background _before_ suggesting it to you, I'm now realizing that it kind of makes the search bar appear disabled. Especially if you look at it on the feature selector. 

Context/apology - a lot of search bars have hover states, but they're usually more subtle and in the form of a drop shadow. Since the WordPress design library isn't big on drop shadows for components, I decided to try using border to indicate hover, which ended up looking goofy. Then I thought using background color would be better. In hindsight, the WordPress design library doesn't have a hover state for the search bar and I guess this could be the reason why 😅 

As terrible as it sounds, I'm leaning toward just taking out the hover state for the search bar, since hover seems to be the most beneficial on clickable components like buttons, dropdowns, etc. I'm so sorry for steering you in the wrong direction 🙇 

> As a middle ground, how would you feel if we applied to the search button the basic hover=neutral-extra-light, active=neutral-light that we do to the rest of our other clear buttons?

Wow, I like this _a lot_! It makes it way clearer that it's a button that can be clicked and I think we should keep that regardless of the search bar hover style 👍

john-legg:
> when the user clears the input field, it might be beneficial to display the initial search state or clear the current results.

I'm so torn on this 😭  On one hand, the current implementation seems useful to allow users to redo their search keywords while being able to reference their previous search results. This is also how Google and Big search results work as well.

However, Google and Bing searches don't debounce the results. I asked GPT and it agrees with Karen's suggestion:
> It is generally better to clear the results when the search box is cleared of search terms. This provides the user with a clear indication that their previous search has been reset, and they can start a new search. However, it's essential to consider the specific use case and user expectations for your product, as there might be scenarios where keeping the previous results would be more user-friendly. 

I agree that it's beneficial to have the search results reflect the search box for consistency, especially since the results instantly update when search terms are manipulated. From a design perspective, I would imagine a cleared search bar would display the initial state with illustration. I could go either way, honestly. Any thoughts, Dan?

> should we implement a minimum character limit before initiating the search, such as 3 characters?
> ...
> I've done things in the past where the debounce (searching after a delay) is restricted to a character limit, but you can always press enter (or the search button) to override that and fire the search anyway. I wonder if that might be a good middle ground?

I think that's perfect, actually! The user will always have a way to "override" the three character limit using those methods. Great suggestion, Dan and Karen 👏

dpasque:
@john-legg @karenroldan sounds good! 👍 

If it's okay with you, I'll put the search clearing behavior in its own PR. That way if we decide later (like in user testing/feedback) that we want to take it back to persisting the results, it's just an easy single revert!

PR #80, opened 2023-04-25
Add feature to allow searching for descriptions

### LINKED ISSUE #58: [Task - Reporting Flow]: Enable searching on descriptions during feature selection
### Details

Use something like an inverted index to match whole words with different weightings. Keep it all on the client. 

pciE2j-1VQ-p2#comment-1674

### Checklist

- [ ]

karenroldan:

This PR updates the implementation for displaying search results in feature selection. Users will now be able to see results when a search term matches a description. 

**Key changes:**
1. Used inverted search only for descriptions. 
2. Filtered the search term with stop words before indexing. 
3. Added tests to cover the changes. 

@john-legg looping you in for design-related! I did not see a Figma for this change, but just checking with you if by simply displaying the tree is okay. 

**Test data:** both the feature and feature group has a description that contains `blog posts`

Screenshot 2023-04-25 at 16 18 22

#### Testing Instructions

* Feel free to explore! Make sure the tests are passing. 

#### Issues

Related to #58 
Closes #

### DIFF
--- a/src/combined-selectors/reporting-config-search-results.ts
+++ b/src/combined-selectors/reporting-config-search-results.ts
@@ -1,10 +1,14 @@
 import { createSelector } from '@reduxjs/toolkit';
 import { includesIgnoringCase } from '../common/lib';
 import { selectFeatureSearchTerm } from '../feature-selector-form/feature-selector-form-slice';
-import { selectNormalizedReportingConfig } from '../static-data/reporting-config/reporting-config-slice';
+import {
+	selectNormalizedReportingConfig,
+	selectIndexedReportingConfig,
+} from '../static-data/reporting-config/reporting-config-slice';
 import {
 	NormalizedReportingConfig,
 	TaskParentEntityType,
+	IndexedReportingConfig,
 } from '../static-data/reporting-config/types';
 import { ReportingConfigSearchResults } from './types';
 import { tokenizeAndNormalize } from '../common/lib';
@@ -12,7 +16,7 @@ import { tokenizeAndNormalize } from '../common/lib';
 function searchReportingConfig(
 	searchTerm: string,
 	reportingConfig: NormalizedReportingConfig,
-	invertedIndex: Map< string, Array< { type: TaskParentEntityType; id: string; weight: number } > >
+	invertedIndex: IndexedReportingConfig
 ): ReportingConfigSearchResults {
 	const { features, featureGroups, products } = reportingConfig;
 	const searchResults: ReportingConfigSearchResults = {
@@ -72,7 +76,7 @@ function searchReportingConfig(
 	const scores: Record< string, { type: TaskParentEntityType; score: number } > = {};
 
 	for ( const token of searchTermTokens ) {
-		const matchingEntities = invertedIndex.get( token ) || [];
+		const matchingEntities = invertedIndex[ token ] || [];
 
 		for ( const { type, id, weight } of matchingEntities ) {
 			if ( ! scores[ id ] ) {
@@ -100,66 +104,13 @@ function searchReportingConfig(
 	return searchResults;
 }
 
-function addTokensToInvertedIndex(
-	entityType: TaskParentEntityType,
-	entityId: string,
-	description: string | undefined,
-	invertedIndex: Map< string, Array< { type: TaskParentEntityType; id: string; weight: number } > >
-) {
-	const descriptionTokens = description ? tokenizeAndNormalize( description ) : [];
-	const tokensWithWeights = [ ...descriptionTokens.map( ( token ) => ( { token, weight: 1 } ) ) ];
-
-	for ( const { token, weight } of tokensWithWeights ) {
-		if ( ! invertedIndex.has( token ) ) {
-			invertedIndex.set( token, [] );
-		}
-
-		const tokenList = invertedIndex.get( token );
-		if ( tokenList ) {
-			tokenList.push( { type: entityType, id: entityId, weight } );
-		}
-	}
-}
-
-function createDescriptionInvertedIndex( reportingConfig: NormalizedReportingConfig ) {
-	const invertedIndex = new Map();
-
-	const { features, featureGroups, products } = reportingConfig;
-
-	for ( const productId in products ) {
-		const product = products[ productId ];
-		addTokensToInvertedIndex( 'product', productId, product.description, invertedIndex );
-	}
-
-	for ( const featureGroupId in featureGroups ) {
-		const featureGroup = featureGroups[ featureGroupId ];
-		addTokensToInvertedIndex(
-			'featureGroup',
-			featureGroupId,
-			featureGroup.description,
-			invertedIndex
-		);
-	}
-
-	for ( const featureId in features ) {
-		const feature = features[ featureId ];
-		addTokensToInvertedIndex( 'feature', featureId, feature.description, invertedIndex );
-	}
-
-	return invertedIndex;
-}
-
 // Searching all of the reporting config is expensive, and these search results are needed by several components.
 // For performance, we need to memo-ize (cache) this piece of derived state.
 // The "createSelector" function lets you  do just that:
 // https://redux.js.org/usage/deriving-data-selectors#optimizing-selectors-with-memoization
 export const selectReportingConfigSearchResults = createSelector(
-	[ selectFeatureSearchTerm, selectNormalizedReportingConfig ],
-	( searchTerm, reportingConfig ) => {
-		return searchReportingConfig(
-			searchTerm,
-			reportingConfig,
-			createDescriptionInvertedIndex( reportingConfig )
-		);
+	[ selectFeatureSearchTerm, selectNormalizedReportingConfig, selectIndexedReportingConfig ],
+	( searchTerm, reportingConfig, indexReportingConfig ) => {
+		return searchReportingConfig( searchTerm, reportingConfig, indexReportingConfig );
 	}
 );
--- a/src/combined-selectors/reporting-config-search-results.ts
+++ b/src/combined-selectors/reporting-config-search-results.ts
@@ -21,6 +21,25 @@ function searchReportingConfig(
 	if ( ! searchTerm ) {
 		return searchResults;
 	}
+
+	const addFeatureGroupAndParents = ( featureGroupId: string ) => {
+		searchResults.featureGroups.add( featureGroupId );
+		searchResults.products.add( featureGroups[ featureGroupId ].productId );
+	};
+
+	const addFeatureAndParents = ( featureId: string ) => {
+		searchResults.features.add( featureId );
+
+		const feature = features[ featureId ];
+		if ( feature.parentType === 'product' ) {
+			searchResults.products.add( feature.parentId );
+		} else {
+			searchResults.featureGroups.add( feature.parentId );
+			const parentFeatureGroup = featureGroups[ feature.parentId ];
+			searchResults.products.add( parentFeatureGroup.productId );
+		}
+	};
+
 	for ( const productId in products ) {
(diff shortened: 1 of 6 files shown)


src/common/lib/string-utils.ts
```diff
@@ -5,3 +5,13 @@ export function includesIgnoringCase( string: string, substring: string ): boole
 export function replaceSpaces( string: string, replacementCharacter = '_' ): string {
 	return string.replaceAll( / /g, replacementCharacter );
 }
+
+export function tokenizeAndNormalize( string: string ) {
+	// Add more stop words as needed
+	const stopWords = new Set( [ 'a', 'an', 'and', 'the', 'in', 'on', 'at', 'of', 'this', 'to' ] );
```
dpasque, line 11: **Praise**: Ahhhh yes!! This is such a great idea, thank you for thinking ahead to this! 😁

src/combined-selectors/reporting-config-search-results.ts
```diff
@@ -54,16 +55,100 @@ function searchReportingConfig(
...
 	return searchResults;
 }
 
+function addTokensToInvertedIndex(
+	entityId: string,
+	description: string | undefined,
+	invertedIndex: Map< string, Array< { id: string; weight: number } > >
+) {
+	const descriptionTokens = description ? tokenizeAndNormalize( description ) : [];
+	const tokensWithWeights = [ ...descriptionTokens.map( ( token ) => ( { token, weight: 1 } ) ) ];
```
dpasque: **Question:** Just for my understanding, is the weight here to add flexibility in case we wanted to weight certain words more in the future? Were there some enhancement ideas you had in mind?

src/combined-selectors/reporting-config-search-results.ts
```diff
@@ -54,16 +55,100 @@ function searchReportingConfig(
...
+				searchResults.products.add( featureGroups[ entityId ].productId );
+			} else if ( features[ entityId ] ) {
+				searchResults.features.add( entityId );
+				if ( features[ entityId ].parentType === 'product' ) {
+					searchResults.products.add( features[ entityId ].parentId );
+				} else {
+					searchResults.featureGroups.add( features[ entityId ].parentId );
+					const parentFeatureGroup = featureGroups[ features[ entityId ].parentId ];
+					searchResults.products.add( parentFeatureGroup.productId );
+				}
```
dpasque: **Nit (non-blocking)**: The logic here for adding an entity and then correctly adding its parents is now repeated in this bigger search function -- how do you feel about making a little inline closure function to DRY it up a bit?

Something like...
```typescript
const addFeatureGroupAndParents( featureGroupId: string ) => {
    	searchResults.featureGroups.add( featureGroupId );
	searchResults.products.add( featureGroups[ featureGroupId ].productId );
}

const addFeatureAndParents( featureId: string ) => {
    searchResults.features.add( featureId );
    if ( features[ featureId ].parentType === 'product' ) {
	    searchResults.products.add( features[ featureId ].parentId );
    } else {
	    searchResults.featureGroups.add( features[ featureId ].parentId );
	    const parentFeatureGroup = featureGroups[ features[ featureId ].parentId ];
	    searchResults.products.add( parentFeatureGroup.productId );
    }
}

// ... then elsewhere ...
if ( products[ entityId ] ) {
    searchResults.products.add( entityId );
} else if ( featureGroups[ entityId ] ) {
    addFeatureGroupAndParents( entityId );
} else if ( features[ entityId ] ) {
    addFeatureAndParents( entityId );
}
```

src/combined-selectors/reporting-config-search-results.ts
```diff
@@ -54,16 +55,100 @@ function searchReportingConfig(
...
+	for ( const { token, weight } of tokensWithWeights ) {
+		if ( ! invertedIndex.has( token ) ) {
+			invertedIndex.set( token, [] );
+		}
+
+		const tokenList = invertedIndex.get( token );
+		if ( tokenList ) {
+			tokenList.push( { id: entityId, weight } );
+		}
+	}
```
dpasque: **Suggestion (blocking)**: (Qualifier: I feel bad because I never documented this assumption anywhere, I'm sorry! 🙇 )

But, to be safe, I think we should avoid assuming that every entity ID is unique across the board. I realize this is currently how things work, but it's something I could see changing in the future. However, we _can_ assume that every entity ID is unique _within_ a given entity map (products, for example) -- that should never change.

All that to say... Would you be okay with storing the id and the entity type here in the index?

Then above, when you are searching the inverted index, you would just check the entity type directly on the result instead of checking for the entity ID in each entity map.

What do you think?

src/combined-selectors/reporting-config-search-results.ts
```diff
@@ -54,16 +55,100 @@ function searchReportingConfig(
...
 // The "createSelector" function lets you  do just that:
 // https://redux.js.org/usage/deriving-data-selectors#optimizing-selectors-with-memoization
 export const selectReportingConfigSearchResults = createSelector(
 	[ selectFeatureSearchTerm, selectNormalizedReportingConfig ],
 	( searchTerm, reportingConfig ) => {
-		return searchReportingConfig( searchTerm, reportingConfig );
+		return searchReportingConfig(
+			searchTerm,
+			reportingConfig,
+			createDescriptionInvertedIndex( reportingConfig )
```
dpasque: **Suggestion (blocking):** Ah, I actually think this may not cache exactly how we need it to. The question of where to do the caching though is kind of a tricky one, so I'll spin up a Slack thread to discuss if that's okay with you! 😄

dpasque (commented):
Karen, this is AWESOME!!! 🎉  This is so cleanly done, and I think it really adds a lot! I really appreciate the future-proofing you've done here too 😄 

I left a few comments, and I'll spin up a Slack thread to talk about the index caching question!

Also, I think @john-legg had some ideas for styling these kind of "soft" matches, so I'll let him weigh in there!

src/combined-selectors/reporting-config-search-results.ts
dpasque: Slack thread: p1682459226372729-slack-CQD1HH4MA

src/combined-selectors/reporting-config-search-results.ts
karenroldan: Good question! I originally implemented this function for names, keywords, and descriptions. We can take the weight out.

src/combined-selectors/reporting-config-search-results.ts
karenroldan: Good catch! I totally missed optimizing this part. Changes are here in [c16cc2e](https://github.com/Automattic/bugomattic/pull/80/commits/c16cc2ee1c13ecd6f89e4151e168ca74deb2b663).

src/combined-selectors/reporting-config-search-results.ts
karenroldan: Oohhh, that makes sense and your suggestion is a good idea! I hope I understood it correctly. The changes can be found here: [109f56b](https://github.com/Automattic/bugomattic/pull/80/commits/109f56bdd44747aabe246ea198a9cd8c25c27d8f)

src/combined-selectors/reporting-config-search-results.ts
```diff
@@ -41,16 +67,37 @@ function searchReportingConfig(
...
+		product: ( entityId: string ) => searchResults.products.add( entityId ),
+		featureGroup: ( entityId: string ) => addFeatureGroupAndParents( entityId ),
+		feature: ( entityId: string ) => addFeatureAndParents( entityId ),
+	};
+
+	for ( const entityId in scores ) {
+		if ( scores[ entityId ].score >= scoreThreshold ) {
+			const entityType = scores[ entityId ].type;
+			const addToSearchResults = addToSearchResultsByType[ entityType ];
+			addToSearchResults( entityId );
```
dpasque: **Praise** Oh this is just a beautiful block of code! 😍  I really like how you grouped the functions to add the entities together here!

src/combined-selectors/reporting-config-search-results.ts
```diff
@@ -41,16 +67,37 @@ function searchReportingConfig(
...
+	const scores: Record< string, { type: TaskParentEntityType; score: number } > = {};
+
+	for ( const token of searchTermTokens ) {
+		const matchingEntities = invertedIndex[ token ] || [];
+
+		for ( const { type, id, weight } of matchingEntities ) {
+			if ( ! scores[ id ] ) {
+				scores[ id ] = { type, score: 0 };
 			}
+			scores[ id ].score += weight;
```
dpasque: **Suggestion:** I might be misunderstanding, but while the block below is now id-safe, I think this part may still not be id-safe? That is, if two entities, across two different types (like a feature and a featureGroup) both had the same ID, they would both get counted in the same score, right? I think we need to make sure the dictionary entries are all unique per entity!

I think there's two ways you go do this:

1) You could make the scores dictionary effectively three dictionaries by type, then just keep using ID:

```typescript
const scores = {
    products: {
        [id]: { } // stuff
    }
    featureGroups: {} // etc
    features: {} // etc
}
```

2) Or you could make the key unique by combining the `type` and `id`! But then you'd also probably want to store the raw ID:

`scores[ type + id ] = { type, id, score: 0 }`

src/static-data/reporting-config/types.ts
```diff
@@ -20,11 +20,9 @@ export interface NormalizedReportingConfig {
 
 /**
  * Indices built from the reporting configuration to facilitate searching for features.
```
dpasque: **Suggestion (non-blocking):** Definitely not needed, but how do you feel about putting a bigger comment here about the shape of the index and what it's used for? I feel like it help might people understand the purpose of this specific index.

src/static-data/reporting-config/reporting-config-parsers.ts
```diff
@@ -316,11 +317,55 @@ function throwReportingConfigError( message: string ): never {
...
+}
+
 export function indexReportingConfig(
-	// eslint-disable-next-line @typescript-eslint/no-unused-vars
 	_response: ReportingConfigApiResponse
 ): IndexedReportingConfig {
-	// TODO: create real implementation once we know the indices we need!
+	const invertedIndex: IndexedReportingConfig = {};
+
+	const { features, featureGroups, products } = normalizeReportingConfig( _response );
```
dpasque: **Suggestion (non-blocking)**: I think we should just change the parameter here to take in a normalized reporting config, so we don't have to normalize it twice! We can totally change the signature here to whatever we need! 👍

dpasque (approved):
The changes look awesome! 😄 I just had a few small comments left, but it shouldn't need any re-review, so approving 👍 

Thank you again for making this change! 🙇

src/static-data/reporting-config/types.ts
karenroldan: Sounds good! Expanded the description here: [343cd60](https://github.com/Automattic/bugomattic/pull/80/commits/343cd609c6989f82de14a04597de52dc974d4a2f)

src/static-data/reporting-config/reporting-config-parsers.ts
karenroldan: Done! Fixed in [a3c7757](https://github.com/Automattic/bugomattic/pull/80/commits/a3c7757c4a76c1d598a0eaea7b2ada31be523d2e)

src/combined-selectors/reporting-config-search-results.ts
karenroldan: Got it! I've changed the implementation here: [3c09993](https://github.com/Automattic/bugomattic/pull/80/commits/3c099936a0db1faa14d76daf32baa8a980c1b22b)

PR #82, opened 2023-04-28
Add re-usable dropdown component API and sort control

### LINKED ISSUE #62: [Task - Duplicate Searching]: Implement the sort filter component
### Details

pciE2j-1VQ-p2#comment-1693

### Checklist

- [ ]

dpasque:

In short, add's the sorting control!

For reference, see: pciE2j-1VQ-p2#comment-1745 (note, we've shifted some of the design slightly since then. Namely, we now just show the exact sort name instead of "Sort" or "Default").

To do that, I've built a re-usable component api: `<Dropdown>`, and its children. See the code comments for more details!
In short though, it's a reusable API for building dropdown style popovers. We will use these in a couple different use cases:
1.  Our custom `select` controls, like this sort filter. 
2. Dropdown menus, like the links to go to the reporting flow and pick the issue type. 

Those behave similarly enough that they can share a lot of the bones!

#### Some a11y notes. 
From an a11y perspective, this sort control is technically a "select-only combobox".
To learn more about comboboxes, you can read up here: https://www.w3.org/WAI/ARIA/apg/patterns/combobox/
This page has the example that is closest to our sort control: https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/

The one thing that is missing from our control is the ability to open the closed control by typing. I tried for a long time to make this work, but I was reallllly fighting the floating-ui API, and it was compromising some of the other important focus behavior. So, for now, typing works once the popover is open, but doesn't open the popover. This is technically still compliant! We can try to enhance this in the future though.

Two weird things you might note:
- For all `comboboxes`, Mac VO likes to say "start typing". There's sadly not much we can do about that! But it shouldn't be a problem, I think the control's behavior is very self-discoverable.
- Mac VO keyboard navigation is different from the native keyboard navigation on the control! Namely, with Mac VO, you always use the left and right keys to navigate among options. But in this control, for compliance, apart from VO you will use the up and down arrows.

### Testing Instructions

- [x] Unit tests pass
- [x] Axe Devtools is clean

#### Issues

Related to #  
Closes #62

### DIFF
--- a/src/common/components/dropdown.tsx
+++ b/src/common/components/dropdown.tsx
@@ -38,8 +38,8 @@ import styles from './dropdown.module.css';
  *   		{Your trigger element}
  * 		</DropdownTrigger>
  * 		<DropdownContent>
- *  		<DropdownItem label="Item 1" role='menuitem'>Item 1</DropdownItem>
- * 			<DropdownItem label="Item 2" role='menuitem'>Item 2</DropdownItem>
+ *  		<DropdownItem typeaheadLabel="Item 1" role='menuitem'>Item 1</DropdownItem>
+ * 			<DropdownItem typeaheadLabel="Item 2" role='menuitem'>Item 2</DropdownItem>
  * 			{... and so on}
  * 		</DropdownContent>
  * </Dropdown>
@@ -225,7 +225,7 @@ interface DropdownItemProps extends HTMLProps< HTMLElement > {
 	/**
 	 * The label is what is used for matching items when typing.
 	 */
-	label: string;
+	typeaheadLabel: string;
 	/**
 	 * The items can be rendered as buttons or links. Default is button.
 	 */
@@ -234,14 +234,14 @@ interface DropdownItemProps extends HTMLProps< HTMLElement > {
 
 export function DropdownItem( {
 	children,
-	label,
+	typeaheadLabel,
 	onClick,
 	className,
 	as: Component = 'button',
 	...props
 }: DropdownItemProps ) {
 	const { setIsDropdownOpen, activeListIndex, getItemProps } = useDropdownContext();
-	const { ref, index } = useListItem( { label } );
+	const { ref, index } = useListItem( { label: typeaheadLabel } );
 
 	const isActive = activeListIndex === index;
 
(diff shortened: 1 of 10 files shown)


src/common/components/dropdown.tsx
```diff
@@ -0,0 +1,275 @@
...
+}
+
+// We're using the React Context API to effectively surface all the FloatingUI data to child components.
+
+type DropdownContext = ReturnType< typeof useDropdown > | null;
+
+const DropdownContext = createContext< DropdownContext >( null );
+
+const useDropdownContext = () => {
+	const context = React.useContext( DropdownContext );
```
karenroldan: **nitpick (non-blocking):** Let's import `useContext` instead to follow the import style for others.

src/duplicate-search/sub-components/sort-select.tsx
```diff
@@ -0,0 +1,93 @@
...
+			label: 'Date added',
+			value: 'date-created',
+		},
+	];
+
+	const sortDropdownTriggerLabel =
+		sortOptions.find( ( sortOption ) => sortOption.value === currentSortOption )?.label || 'Sort';
+
+	const handleSortOptionClick = useCallback(
+		( sortOptionValue: string ) => {
```
karenroldan: **suggestion (non-blocking):** I think it will be better and safer if we use `IssueSortOption` type as the parameter instead of string!  

I see that parameter is cast when passing to `setSort`, it'll be better to enforce the correct type in the beginning.

karenroldan (approved):
Great work on these changes! 👏🏻  it tests really nicely for me and I like the idea of having the type-ahead searching which is a great touch for usability! 

I added a couple of non-blocking comments.

john-legg (approved):
Great job on this PR, Dan 👏 I know it was tough one, but I think you really nailed it 🔨 

For the first time, I don't have any comments 😆 The functionality with VoiceOver worked perfectly and I honestly wouldn't have known that there were any compromises! The typeahead feature is cool too even if it's only when the dropdown is open. Nicely done 🚀

src/common/components/dropdown.tsx
dpasque: Whoops, nice catch! 😄 Got it updated 👍

src/duplicate-search/sub-components/sort-select.tsx
dpasque: Good call, got it fixed! 👍

PR #83, opened 2023-04-28
Fix search handling

dpasque:

This adds three main fixes to our debounced search handling:
1. If a search term is the exact same as the last emitted search term, we won't debounce it. However, if users still hit enter or the search button, we will emit it. Think of typing "abc", then after the debounce, quickly deleting and re-adding the "c".
2. We remove the forced searching on input blur (lose focus). That was way too eager lol. 
3. If the user clears the search term on the duplicate search page we show the placeholder.

You can basically think of our searching rules as: If the user ever hits enter or the search button, always respond to the behavior. In the case of debouncing, we add optimizations, etc.

One note re: a11y -- we don't trigger any screen reader alerts when the user clears their search. That is intentional! Nothing is actually changing meaningfully in the results (no search is actually firing), so we should avoid spamming the user with alerts.

#### Testing Instructions

- [x] Unit tests pass

Feel free to play around with the duplicate search bar and the feature selector! 👍 

#### Issues

Related to #  
Closes #

### DIFF
--- a/src/common/components/debounced-search.tsx
+++ b/src/common/components/debounced-search.tsx
@@ -49,7 +49,11 @@ export function DebouncedSearch( {
 
 	const searchImmediately = () => {
 		debouncedCallback.cancel();
-		callCallback( searchTerm );
+		const trimmedSearchTerm = searchTerm.trim();
+		// See the comment below on trimming -- after hitting enter, the user would expect to
+		// see the trimmed input in the actual input field.
+		setSearchTerm( trimmedSearchTerm );
+		callCallback( trimmedSearchTerm );
 	};
 
 	const handleEnter = ( event: React.KeyboardEvent ) => {
@@ -61,7 +65,15 @@ export function DebouncedSearch( {
 	const handleChange = ( event: React.ChangeEvent< HTMLInputElement > ) => {
 		const newSearchTerm = event.target.value;
 		setSearchTerm( newSearchTerm );
-		debouncedCallback( newSearchTerm );
+		debouncedCallback( newSearchTerm.trim() );
+	};
+
+	// We only ever call the callback with a trimmed search term.
+	// However, trimming the actual value in the input on change, or even on debouncing
+	// feels really weird and clunky to the user. So we only trim the actual value in the
+	// input field on enter (or search button) and on blur.
+	const handleBlur = () => {
+		setSearchTerm( searchTerm.trim() );
 	};
 
 	const classNames = [ styles.search ];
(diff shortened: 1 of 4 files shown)


src/common/components/debounced-search.tsx
```diff
@@ -23,14 +23,33 @@ export function DebouncedSearch( {
...
-		() => debounce( callback, debounceMs ),
-		[ callback, debounceMs ]
+		() => debounce( conditionallyCallCallback, debounceMs ),
+		[ conditionallyCallCallback, debounceMs ]
 	);
 
 	const searchImmediately = () => {
 		debouncedCallback.cancel();
-		callback( searchTerm );
+		callCallback( searchTerm );
```
karenroldan: **suggestion (non-blocking):** What do you think about trimming the search term so it will remove any leading or trailing whitespace first? 

When a user enters only space and presses enter, we should show the placeholder and treat it as an empty search term instead of showing the results for the previous search term. 

**Steps to reproduce:** 

1. Search for the term "test".
2. Wait for the search results. 
3. Clear the search input.
4. Press enter and see the placeholder displayed as expected.
5. Enter a single space in the search input and press enter.
6. Notice that the search results for the previous term "test" are displayed instead of treating it as an empty search term.

karenroldan (approved):
Yay! These changes are exciting! Overall, it works really great. 🚀

src/common/components/debounced-search.tsx
john-legg: Wow, that's a great catch, Karen! 👏

john-legg (approved):
Nicely done, Dan! The search feels _really_ smooth, even on VoiceOver! This makes two PRs in a row that I haven't had anything to comment on 😆 

> We remove the forced searching on input blur (lose focus). That was way too eager lol.

Awesome! Yeah, this was a bit jarring, especially if you clicked on a filter or sort dropdown immediately after search, causing it to go out of focus and debounce again.

src/common/components/debounced-search.tsx
dpasque: Fantastic idea!! Got it implemented 👍 

There are comments to the effect, but the short of the trimming behavior is...
- We only ever emit trimmed searches
- On enter/search button, we change the input display to show the trimmed term
- On blur, we change the input display to show the trimmed term
- Notably, we don't change the actual input display to the trimmed term on debouncing only. This felt kind of jarring -- the cursor would just jump all over the place! 😆

PR #87, opened 2023-05-02
(Real, recovered) Add page transition and navigations

### LINKED ISSUE #64: [Task - Duplicate Searching]: Create buttons/links to navigate between two app pages
### Details

pciE2j-1VQ-p2#comment-1693

Also should include the banner at the bottom. 

### Checklist

- [ ]

dpasque:

Gahhh, sorry this PR got so big! 🙇 A lot of these page transitions/navigations changes all go together, so it was hard to break it up without creating a big rebase chain. And this change also prompted some needed modularizations to DRY things up!

But all in all, the changes are smaller than what may first appear!

First, some modularizations:
- Creating a new standardize `Banner` component for the banners (report a new issue, and start over) at the bottom of both pages.
- Creating a re-usable `ReportIssueDropdownMenu` that can be used in both the top navbar and the report a new issue banner.
- Moving the determination of what the next step in the reporting flow into a shared selector for better reuse.
- Creating a `PageNavigationProvider` and a `usePageNavigation` hook that takes care of doing all the side-effects when we change pages (like focusing the header, scrolling to the top, updating the title).

With that, we could finish everything up!
- Adding a real navbar to the top of the site
- Adding the "report an issue" banner at the bottom of the duplicate searching page
- Updating the "start over" banner at the bottom of the reporting flow page

#### Testing Instructions

- [ ] Unit tests pass
- [ ] Axe devtools is clean

A quick heads-up: there is handling in there to not focus on the page subheading (like "Report a new issue") when we first render the app. However, we use `React.StrictMode` to catch errors while developing, and `StrictMode` famously renders everything twice. So, that behavior won't work quite right locally. But it will in production! If you want to see for yourself, you can just delete the `StrictMode` wrappers in `index.tsx` 🙂 .

#### Issues

Related to #  
Closes #64


karenroldan (commented):
AAH, the changes look really good! Bugomattic is hereeee! 👏🏻 

- There's a nice touch on the menu items and switching between pages is seamless. 
- I like how the issue type is already set after selecting from the menu. 
- The navigation is really intuitive! It's nice how it always focuses on the top when you navigate and doesn't refresh the entire page! 
- Overall, it looks great both on desktop and mobile! 

Really well done on these changes 👏🏻 🥳 

NOTE: I'm halfway through with the review. I'll go over the rest later!

dpasque:
Ahhh thank you @karenroldan! 😄 Thank you for taking on this monster of a PR, and I'm so sorry it got so big haha...

src/app/__tests__/app.test.tsx
```diff
@@ -96,4 +96,19 @@ describe( '[app]', () => {
 		);
 		expect( monitoringClient.analytics.recordEvent ).toHaveBeenCalledWith( 'page_view' );
 	} );
+
+	// This main seem silly, but this is an important test!
+	// We set focus on page headers on page navigation, and have to make sure we don't do that on initial load.
+	// This test verifies that! The page heading focus is tested elsewhere.
+	test( 'On app boot, focus is on the body element', async () => {
```
karenroldan, line 103: **praise:** TIL! Thanks for adding comments for reference.

karenroldan (approved):
Continuing from my previous comment! Overall, the code looks great -- I mainly focused on its readability and maintainability. I'm always impressed with how neatly-structured code can be written. I also noticed that you included comments throughout the code, which is really helpful!

I noticed only two things from my testing: 

1. When using the tab key for navigation, it's not possible to access the "Report an Issue" dropdown. The arrow keys do function correctly, however! 
2. VoiceOver: When navigating to the Reporting Flow page after selecting an issue type, VoiceOver only announces "heading level 2, Report a new issue, main". It would be helpful to provide more context such as announcing the specific form that is currently open. 

Great work as usual! 🎉

john-legg (approved):
This is great, Dan! I think you did an amazing job on the implementation 👏 VoiceOver works just as you'd expect and it even looks good on mobile.

**nitpick (non-blocking):**
Apologies for not giving you a heads up, but the Figma file was updated a few weeks ago with a different hover and clicked state for the tabs, which is more aligned with the WordPress design library.
Screenshot 2023-05-08 at 5 38 39 PM

I originally had the background hover and clicked state match other components, which is what you might have seen. This was when I was on my "consistency no matter what" mission 😆 Seeing it live, the background change is a bit distracting although it does help make it clear that it's clickable.

If there's no accessibility concerns, would you mind updating the hover and click states? The focused state is fine by the way!

dpasque:
@karenroldan Thank you for the review and the feedback! 😄 

> When using the tab key for navigation, it's not possible to access the "Report an Issue" dropdown. The arrow keys do function correctly, however!

This may sound and feel a little weird initially, but this is actually per WAI ARIA specs! Essentially, you're only supposed to be able to tab into one piece of a `menubar`, and then use keyboards to move within then. We are following a "roving tabindex" pattern to make that tabable element whichever one was last focused! 👍 

> VoiceOver: When navigating to the Reporting Flow page after selecting an issue type, VoiceOver only announces "heading level 2, Report a new issue, main". It would be helpful to provide more context such as announcing the specific form that is currently open.

Oh I love this idea! 😄  I added a screen-reader-only span that is also set as the accessible description of the `h2` heading! 🎉

dpasque:
@john-legg Ah shoot, sorry I missed that! Thanks for the heads up! 😄 

Got those styles fixed 👍  Thank you again for the review! 🙌

PR #89, opened 2023-05-04
Add error handling/display to duplicate search

### LINKED ISSUE #78: [Task - Duplicate Searching]: Handle an errored search request
### Details

We should come up with something to show when searching fails

### Checklist

- [ ]

dpasque:

We previously didn't handle any request errors when duplicate searching. Now we do!

We show a little error display and log the error (with caching to prevent spam).

#### Testing Instructions

- [x] Unit tests pass

Feel free to play around locally! Just modify the `local-api-client` to throw an error when searching issues. 

#### Issues

Related to #  
Closes #78

### DIFF
--- a/src/duplicate-results/duplicate-results.tsx
+++ b/src/duplicate-results/duplicate-results.tsx
@@ -4,22 +4,29 @@ import {
 	selectDuplicateResults,
 	selectDuplicateResultsRequestStatus,
 	selectDuplicateRequestsWereMade,
+	selectDuplicateResultsRequestError,
 } from './duplicate-results-slice';
 import { IssueList, PlaceholderMessage, useShowBanner } from './sub-components';
 import styles from './duplicate-results.module.css';
 import { ReactComponent as InitialIllustration } from './svgs/initial-illustration.svg';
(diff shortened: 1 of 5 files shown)


src/duplicate-results/duplicate-results.tsx
```diff
@@ -41,7 +52,7 @@ export function DuplicateResults() {
 	if ( ! requestsWereMade || searchTerm === '' ) {
 		resultsContainerDisplay = (
 			<PlaceholderMessage
-				illustration={ InitialIllustration }
+				illustration={ <InitialIllustration /> }
```
karenroldan, line 55: **question (non-blocking):** Out of curiosity, is there a particular reason to pass the component itself rather than the rendered instance?

karenroldan (approved):
LGTM! 🚀

src/duplicate-results/duplicate-results.tsx
dpasque: Ah good question! AFAIK, I don't think it makes too much a difference, but I'm not entirely sure... This case was more about consistency and getting the typing to play together nicely. I wanted to apply a class upriver to one of the illustration components, so that was the motivation to switching over to this syntax!

PR #92, opened 2023-05-05
Display description & keyword matches in a pill

karenroldan:

This PR now displays the description matches! 🎉  

- If the search term matches the description or keyword, a **pill with a generic text** will be displayed.
- If there are matches on the name and keyword/description, we don't display the pill and will highlight the matched name. 
- If there are matches on both the keyword and description, we only display the matched keyword. 

**Key changes:**

1. Updated the whole `reporting-config-search-results.tsx` and separated it into more readable functions. 
2. Match type is now stored directly in the search results 
3. Hierarchy of matches (name > keyword > description) is now computed in the search results 
4. On mobile, only the icon will be displayed

Screenshot 2023-05-12 at 09 38 35
Screenshot 2023-05-12 at 09 38 08
Screenshot 2023-05-12 at 07 29 00

Screenshot 2023-05-12 at 07 28 50

#### Testing Instructions

* I've added and updated the tests to cover all possible scenarios I could think of. Feel free to explore! 
* All tests should pass. 

#### Issues

Related to #  
Closes #


src/combined-selectors/reporting-config-search-results.ts
```diff
@@ -99,15 +116,61 @@ function searchReportingConfig(
...
+	if ( ! searchTerm ) {
+		return searchResults;
+	}
+
+	searchByName( searchTerm, products, featureGroups, features, searchResults );
+
+	const searchTermTokens = tokenizeAndNormalize( searchTerm );
+	searchInDescription( searchTermTokens, invertedIndex, searchResults, featureGroups, features );
+
+	findStrongestMatchInDescription( searchResults );
```
dpasque: **Suggestion (non-blocking)**:

I really like how you've broken up the searching steps here -- I think it cleans it up a lot and makes it a lot more modular! 🎉 

In reading it, having so many arguments to the functions slowed me down just a little bit. I think the tricky thing here is that we have a kind of complicated search that is inherently kind of stateful, and we have to pass all that state around!

From that perspective, what do you think of potentially moving this to a [builder pattern](https://refactoring.guru/design-patterns/builder)?

Definitely not necessary, and it's really a matter of preference, so I'll defer the ultimate decision up to you! (Really, no hard feelings if you'd rather not go this way! 😁 )

I tried it out locally real quick (all the same logic, just as a builder) and will paste that below so you can decide if you like it or not!

```typescript
class ReportingConfigSearcher {
	private searchResults: ReportingConfigSearchResults;
	private products: Products;
	private featureGroups: FeatureGroups;
	private features: Features;
	private invertedIndex: IndexedReportingConfig;
	private searchTerm: string;

	constructor(
		searchTerm: string,
		reportingConfig: NormalizedReportingConfig,
		invertedIndex: IndexedReportingConfig
	) {
		const { products, featureGroups, features } = reportingConfig;
		this.products = products;
		this.featureGroups = featureGroups;
		this.features = features;

		this.invertedIndex = invertedIndex;

		this.searchResults = {
			products: new Set(),
			featureGroups: new Set(),
			features: new Set(),
			descriptionMatchedTerms: {},
			strongestMatch: { product: {}, featureGroup: {}, feature: {} },
		};

		this.searchTerm = searchTerm;
	}

	private addFeatureGroupAndParents( featureGroupId: string ): void {
		this.searchResults.featureGroups.add( featureGroupId );
		this.searchResults.products.add( this.featureGroups[ featureGroupId ].productId );
	}

	private addFeatureAndParents( featureId: string ) {
		this.searchResults.features.add( featureId );

		const feature = this.features[ featureId ];
		if ( feature.parentType === 'product' ) {
			this.searchResults.products.add( feature.parentId );
		} else {
			this.searchResults.featureGroups.add( feature.parentId );
			const parentFeatureGroup = this.featureGroups[ feature.parentId ];
			this.searchResults.products.add( parentFeatureGroup.productId );
		}
	}

	searchByName() {
		for ( const productId in this.products ) {
			const product = this.products[ productId ];
			if ( includesIgnoringCase( product.name, this.searchTerm ) ) {
				this.searchResults.products.add( productId );
			}
		}

		for ( const featureGroupId in this.featureGroups ) {
			const featureGroup = this.featureGroups[ featureGroupId ];
			if ( includesIgnoringCase( featureGroup.name, this.searchTerm ) ) {
				this.addFeatureGroupAndParents( featureGroupId );
			}
		}

		for ( const featureId in this.features ) {
			const feature = this.features[ featureId ];
			const keywordsIncludeSearchTerm = () =>
				feature.keywords?.some( ( keyword: string ) =>
					includesIgnoringCase( keyword, this.searchTerm )
				);

			if ( includesIgnoringCase( feature.name, this.searchTerm ) || keywordsIncludeSearchTerm() ) {
				this.addFeatureAndParents( featureId );
			}
		}

		return this;
	}

	searchInDescription() {
		const searchTermTokens = tokenizeAndNormalize( this.searchTerm );
		const scores = {
			product: {} as Record< string, number >,
			featureGroup: {} as Record< string, number >,
			feature: {} as Record< string, number >,
		};

		for ( const token of searchTermTokens ) {
			const matchingEntities = this.invertedIndex[ token ] || [];

			for ( const { type, id, weight } of matchingEntities ) {
				if ( ! scores[ type ][ id ] ) {
					scores[ type ][ id ] = 0;
				}
				scores[ type ][ id ] += weight;

				// Store the matched terms in the matchedEntities object
				if ( ! this.searchResults.descriptionMatchedTerms[ type ] ) {
					this.searchResults.descriptionMatchedTerms[ type ] = {};
				}
				if ( ! this.searchResults.descriptionMatchedTerms[ type ][ id ] ) {
					this.searchResults.descriptionMatchedTerms[ type ][ id ] = new Set();
				}
				this.searchResults.descriptionMatchedTerms[ type ][ id ].add( token );
			}
		}
		const scoreThreshold = 1;

		for ( const entityId in scores.product ) {
			if ( scores.product[ entityId ] >= scoreThreshold ) {
				this.searchResults.products.add( entityId );
			}
		}

		for ( const entityId in scores.featureGroup ) {
			if ( scores.featureGroup[ entityId ] >= scoreThreshold ) {
				this.addFeatureGroupAndParents( entityId );
			}
		}

		for ( const entityId in scores.feature ) {
			if ( scores.feature[ entityId ] >= scoreThreshold ) {
				this.addFeatureAndParents( entityId );
			}
		}

		return this;
	}

	findStrongestMatchInDescription() {
		const descriptionMatches = this.searchResults.descriptionMatchedTerms;
		for ( const entityType in descriptionMatches ) {
			for ( const entityId in descriptionMatches[ entityType ] ) {
				const matchedTerms = descriptionMatches[ entityType ][ entityId ];
				let strongestMatch = '';
				let maxScore = 0;
				for ( const term of matchedTerms ) {
					const currentScore = descriptionMatches[ entityType ][ entityId ].size;

					if ( currentScore > maxScore ) {
						strongestMatch = term;
						maxScore = currentScore;
					}
				}
				this.searchResults.strongestMatch[ entityType ][ entityId ] = strongestMatch;
			}
		}

		return this;
	}

	getSearchResults() {
		return this.searchResults;
	}
}

function searchReportingConfig(
	searchTerm: string,
	reportingConfig: NormalizedReportingConfig,
	invertedIndex: IndexedReportingConfig
): ReportingConfigSearchResults {
	const searcher = new ReportingConfigSearcher( searchTerm, reportingConfig, invertedIndex );

	if ( ! searchTerm ) {
		return searcher.getSearchResults();
	}

	return searcher
		.searchByName()
		.searchInDescription()
		.findStrongestMatchInDescription()
		.getSearchResults();
}
```

src/feature-selector-form/sub-components/matched-terms-display.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+	return (
+		<span className={ styles.termMatchesWrapper }>
+			<Pill key={ matchType } highlightClassName={ styles.matchedPill }>
+				<MatchIcon data-testid="match-icon" aria-hidden={ true } className={ styles.appIcon } />
+				<TextMatchHighlighter
+					textMatch={ searchTermRegExp }
+					highlightClassName={ styles.searchSubstringMatch }
+				>
+					{ `${ matchType } match: ${ matches }` }
```
dpasque: **Bug (blocking)**: As a heads up, because all of this string is in the `TextMatchHighlighter`, this means that if the description contains the word "description" or "match", that prefix will also highlight!
Screen Shot 2023-05-09 at 1 34 40 PM

src/feature-selector-form/sub-components/feature-group.tsx
```diff
@@ -14,20 +20,41 @@ interface Props {
...
+				<MatchedTermsDisplay
+					searchTerm={ searchTerm }
+					matchedTerms={ matchedDescriptionTerms }
+					strongestMatchTerm={ strongestDescriptionMatch }
+					matchType={ 'description' }
+				/>
+			</>
+		);
+	}
+
```
dpasque: **Suggestion (non-blocking)** This suggestion has a few moving pieces, but I'll leave the comment here!

I was wondering if there might be a way to simplify all the search match display logic, and centralize (DRY) it up even more.

I'm wondering if the key to that is two-fold...
1. When searching, start storing the match type directly on the search result (instead of separate properties for the IDs and then the maps that point IDs to description terms, etc.)
2. Let the entire label display for each tree node be controlled by the `MatchedTermsDisplay` component (maybe with a slightly different name).


So, for 1) above, that just means that when we're doing the searching, we build an object instead that looks like this:

```typescript
interface NameMatch {
	matchType: 'name';
}

interface KeywordMatch {
	matchType: 'keyword';
	keyword: string;
}

interface DescriptionMatch {
	matchType: 'description';
	matchedTerms: Set< string >;
	strongestMatch: string;
}

interface Matches {
	[ entityId: string ]: NameMatch | KeywordMatch | DescriptionMatch;
}

export interface ReportingConfigSearchResults {
	products: Matches;
	featureGroups: Matches;
	features: Matches;
}
```

Then for 2), we make `MatchedTermsDisplay` just take in the entity ID and type, and it handles comparing that with all the search info (search term and results), and building the entire display.

That would centralize the "hierarchy" of preferring name matches > keyword matches > description matches. It would also probably fix the bug where you can show both a name match and a description match at the same time:

Screen Shot 2023-05-09 at 1 35 23 PM

And it would simplify the code for the `Product`, `FeatureGroup`, and `FeatureComponent`, because it would just look like:
```typescript
const label = <MatchedTermsDisplay entityId={ id } entityType={ 'featureGroup' } />

return (
<ExpandableTreeNode
	label={ label }
	label={ labelDisplay }
	isExpanded={ isExpanded }
	handleToggle={ handleCollapseExpandToggle }
	description={ description }
>
	<SortedFeatureList
		featureIds={ featureIdsToDisplay }
		parentName={ name }
		treeLevel={ 'third' }
	/>
</ExpandableTreeNode>
);
```

Again, not required, so I'll defer to you if you like it better! But I think it might read a pinch more centralized/simplified

dpasque (commented):
Awesome, thank you @karenroldan! 🎉  I think this is off to a really solid start! This is a really complex area, and I think you've done a nice job handling that. 😄 

I left a few comments with some refactoring ideas that might streamline it even more -- feel free to build off those ideas if you like them! No pressure to take them though, of course 🙂 

I found a few small bugs that I mention in some of those other comments, but I thought I would list here:

1. If the description contains the word "description" or "match", we highlight that in the prefix:
Screen Shot 2023-05-09 at 1 34 40 PM

2. If there's a match in the name and description, we show the description match. I believe we should just show the name in this case:
Screen Shot 2023-05-09 at 1 35 23 PM

3. It looks like the pill expands the height of the tree row slightly compared to rows without pills. Would there be a way to keep that consistent so all the tree rows stay the same height?

Finally, I wanted to have a quick design discussion after playing around with that and building off of: p1682618610780909-slack-CQD1HH4MA. (bringing in @john-legg )

Right now, the design still feels just a little busy to me, and I'm also wondering if there's anyway we can minimize the overall complexity just for the sake of future maintenance!

I feel so bad asking this after @karenroldan did such a nice job implementing it, but I'm wondering if we should literally just display the phrase "description match" in the pill without showing any keywords or scores. This is less information, but I think it makes the tree look less overwhelming, prevents wrapping in most cases, and also lets us simplify a lot of the code!

I also wonder what is the least bit of information we can show for keyword matches. Right now we have the search icon, and `keyword match: {keyword}`. Do we even need the `keyword-match` prefix? Could we just re-use the blue pill color that we use in the bottom panel? 

Let me know what y'all think! 😄

dpasque:
I'm also realizing that a lot of these suggestions (both design and code refactor) might need further discussion or clarification, so please just feel free to spin up a Slack thread if it would be more helpful to discuss these! 🙂 👍

john-legg:
> I'm also realizing that a lot of these suggestions (both design and code refactor) might need further discussion or clarification, so please just feel free to spin up a Slack thread if it would be more helpful to discuss these! 🙂 👍

Yeah, a Slack thread might be a good idea! I'll go ahead and start one 👍

karenroldan:
> I feel so bad asking this after @karenroldan did such a nice job implementing it, but I'm wondering if we should literally just display the phrase "description match" in the pill without showing any keywords or scores. This is less information, but I think it makes the tree look less overwhelming, prevents wrapping in most cases, and also lets us simplify a lot of the code!

Don't worry about it! I like that we can keep on iterating and explore on different ideas. Let's continue to weigh the pros and cons and find the best solution! 💡 

We can continue our discussion in this Slack thread: p1683670129349769-slack-CQD1HH4MA

src/combined-selectors/reporting-config-search-results.ts
karenroldan: > From that perspective, what do you think of potentially moving this to a [builder pattern](https://refactoring.guru/design-patterns/builder)?

OHH, I definitely like this better! Thanks for sharing the codes also! 🙇🏻‍♀️ changes can be found here [baa3bc3](https://github.com/Automattic/bugomattic/pull/92/commits/baa3bc3aafee400e5b66705df20f5375ed318894)

src/feature-selector-form/sub-components/matched-terms-display.tsx
karenroldan: Good catch! I moved the label outside the `TextMatchHighlighter`. 
Screenshot 2023-05-10 at 11 47 04

src/combined-selectors/types.ts
```diff
@@ -1,5 +1,29 @@
+export type MatchType = 'name' | 'keyword' | 'description' | 'child';
+export type MatchOption = NameMatch | KeywordMatch | DescriptionMatch | ChildMatch;
+
+interface ChildMatch {
```
karenroldan: When the search term matches a feature in the lowest level of the hierarchy, we add its parents to the search results. BUT, we don't want to indicate that the parents themselves match the description. Instead, we include them in the search results to display the hierarchical relationship in the tree.

src/combined-selectors/reporting-config-search-results.ts
```diff
@@ -8,108 +8,181 @@ import {
...
+
+	private isPriorityMatch( existingMatchType: MatchType, newMatchType: MatchType ): boolean {
+		const priority = {
+			name: 3,
+			keyword: 2,
+			description: 1,
+			child: 0,
+		};
+		return priority[ newMatchType ] >= priority[ existingMatchType ];
+	}
```
dpasque: **Praise**: Your whole typing system here, and the prioritization scheme, reads so nicely! I love this approach! 🎉

src/combined-selectors/reporting-config-search-results.ts
```diff
@@ -8,108 +8,181 @@ import {
...
+				scores[ type ][ id ] += weight;
+			}
+		}
 
-	for ( const token of searchTermTokens ) {
-		const matchingEntities = invertedIndex[ token ] || [];
+		const scoreThreshold = 1;
+		const descriptionMatch: () => MatchOption = () => ( {
+			matchType: 'description',
+			matchedTerms: new Set( searchTermTokens ),
```
dpasque: **Suggestion (non-blocking)**: Hmm, I think this is always going to be just all the tokens in the search phrase, right? So I don't think this will actually be the unique matched terms.

But looking at this, it looks like we don't actually use this field right now! Maybe we can just pull it for now, and then we can add it back to the type later if we need it? 👍 

I'll make that update

dpasque (approved):
This looks sooooooo good! 😄 Nice work, and thanks for being so flexible and making all the design updates!

I pushed a couple of tiny CSS and typing tweaks, I'll go ahead and merge this if that's okay with you!

Issue #93, opened 2023-05-10
[Bug]: Duplicate search term doesn't sync between input and redux state when using URL history


dpasque:
### Quick summary

The duplicate search term doesn't stay in sync between the redux and input when parsing from the URL history.

You see two main bugs, both with the same root cause:
1. If you load from a URL, the search term is set in the redux store, but not he actual search input
2. When you use forward/back, we change the redux state, but not the input value

Both are because the local state of the debounced search is completely decoupled from the redux values.

### Steps to reproduce

1. Add a search term and search
2. Refresh the page with the new URL -- no search term!

### Other notes

What we need to do is...
1. Make the `DebouncedSearch` optionally take in "controlled" state values (`searchTerm`, `setSearchTerm`) in the props. If we provide those, use those instead of the local state values. 
2. Take the parent duplicate search component and sync the local state with the redux value

PR #94, opened 2023-05-10
Make the URL cleaner and more user friendly

### LINKED ISSUE #69: [Task - Other]: Don't store nullish/empty values in the URL
### Details

Right now, the minute you do a history updating action in the site, you get this giant URL!

This is because we are taking all of the tracked state and serializing it in the URL. However, we don't need to serialize all of it all of the time. If things have nullish values (like empty arrays, empty strings, null, undefined), we can just leave that property out of the URL. This will make the URL simpler, and more inviting for people to punch values in or bookmark, etc.

In short, everything in the URL will be meaningful!

### Checklist

- [ ] Update the serializer to exclude nullish/empty values
- [ ] Make sure all the parsers in the reducers handle these okay!

dpasque:

Don't ya just hate when you go to Bugomattic, make one teeny, innocent change, and then get steamrolled by a URL that looks like the green symbols from the Matrix? I sure do!

It's time to make our state-tracking URL more user-friendly!

**Note, these changes are actually quite small. I just updated some static string values, and that corresponded to a lot of file updates!

Here's how our URL is now better:
- We don't track falsy values or empty arrays.
- For more parameter-like state, we don't track the value if it's the default. The notable exception are the `activePage` and the `activeReportingStep`. (These feel more like navigation values, so hiding the default felt weird. Although I feel that way much more so about `activeStep` than `activeReportingStep`. What do you think?)
- For objects, we use "dot" notation instead of brackets. Fullstops don't need to be encoded, so I think this reads wayyyyy nicer.
- I simplified the values for all our pre-defined string types (like `IssueType` and `ActivePage`) and switched from camelCase values to kebab-case values. I think that will be more user-friendly for manual editing!
- I changed the order of the state keys in the URL

#### Testing Instructions

- Unit tests still pass
- Navigate through the app, and see how the URL now looks

Note: the search term appears to be borked, and was so before this change. I've written up the issue in #93 and will fix it in another PR 👍 

#### Issues

Related to #  
Closes #69

### DIFF
--- a/src/url-history/parsers.ts
+++ b/src/url-history/parsers.ts
@@ -23,10 +23,32 @@ export function stateToQuery( state: RootState ) {
 	for ( const key of trackedStateKeys ) {
 		stateToSerialize[ key ] = state[ key ];
 	}
-	const query = qs.stringify( stateToSerialize );
+	const query = qs.stringify( stateToSerialize, {
+		filter( prefix, value ) {
+			if ( isFalsyOrEmpty( value ) ) {
+				return;
+			}
+
+			if ( defaultStateValues[ prefix ] === value ) {
+				return;
+			}
+
+			return value;
+		},
+	} );
 
 	return query;
 }
+
+function isFalsyOrEmpty( value: unknown ) {
+	return ! value || ( Array.isArray( value ) && value.length === 0 );
+}
+
+const defaultStateValues: { [ key: string ]: string } = {
+	'duplicateSearch[statusFilter]': 'all',
+	'duplicateSearch[sort]': 'relevance',
+	'issueDetails[issueType]': 'unset',
+};
 
 export function queryToState( query: string ): Partial< RootState > {
 	const queryObject = qs.parse( query, {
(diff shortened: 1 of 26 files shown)


src/url-history/parsers.ts
```diff
@@ -7,11 +7,12 @@ import qs from 'qs';
 type KeyableRootState = Omit< RootState, keyof EmptyObject >;
 
 // If you want any redux state to be tracked in the URL, add the top level key here.
```
dpasque, line 9: As an FYI: this is where all the meaningful code changes are! Everything else is just string type updates 👍

worldomonation (commented):
I don't know anywhere near enough React, but the changes in this PR makes sense even to someone like me.

You've gone from camelCase URL to a dash-case(?) format instead. I think that does make it easier to read in the context of URLs.

dpasque:
@worldomonation  Awesome! 😄 Thanks for taking a look!!

> You've gone from camelCase URL to a dash-case(?) format instead. I think that does make it easier to read in the context of URLs.

Yep, that's a big change! Probably the biggest one though is not tracking the falsy or default values -- mostly this just keeps the number of query params appended to the end to a much smaller number except in really complicated case. I think this will help people feel more comfortable learning all the query params! And it makes conceptual sense: as you add meaningful data to the app, the URL gets correspondingly longer and more data rich (rather than just all at once)

karenroldan (approved):
I like how the URL looks! It's more readable and the structure is easier to understand now 🎉

PR #95, opened 2023-05-10
Making searching and URL history tracking play together nicely

### LINKED ISSUE #93: [Bug]: Duplicate search term doesn't sync between input and redux state when using URL history

dpasque:

Previously, our URL history state tracking and our duplicate searching controls did not work together very well.

There were two main bugs:
1. The actual value in the search `input` was decoupled from the redux search value. So if you navigated back/forth, the value you saw wouldn't change, even though the state was changing. Same was true on initial load if a search term came in from the URL.
2. We didn't actually update any of the search results! So all your search parameters would change, but all the results would be from old searches!

This PR touches a lot of files, but the core changes to fix the above are...
1. The `DebouncedSearch` optionally lets you "control" the `input` local state, i.e., define it in the parent component and pass it down. This now lets us sync the local `input` state with the search term as stored in the redux state!
2. We've moved our duplicate searching from an imperative approach (i.e., you click a button, and we dispatch a search thunk), to a reactive approach (if we see a search parameter change in the redux store, we fire a new search). We use redux middleware to implement this reactive approach because it is completely decoupled from the React render cycles (unlike `useEffect`). This effectively means that navigating forward and back in the browser, and doing an initial load with search parameters, will all kick off a new search, keeping the search results in sync with the search parameters. 🎉 
3. Because we now might be repeating a lot of similar searches (especially in the case of forward/back navigation), we add an in-memory cache for search requests. That makes the search result updates appear near instantaneously when using forward/back navigation.

#### Testing Instructions

I added a lotttt more component tests covering all kinds of permutations of history + duplicate searching! 

For playing around manually, it might not be a bad idea to tweak the `local-api-client.ts` file and add a `console.log` to the search request. With caching now, it can be hard to tell when a request is firing!

The rules that govern when you should searches fire can feel a bit convoluted, so I'll lay them out here:
1. Never fire a search if the search term is empty.
2. If the user takes an action that dispatches an action that updates the search state in the redux store, fire a search. This is even true if the search parameters are the same! Put another way, hitting "enter" on the same search term will search again, albeit with a cache, so it may not _feel_ like it.
3. If the URL history changes (forward/back), only fire a search if the values of the search parameters have changed. That way, browser navigations that have nothing to do with duplicate searching aren't just spamming pointless search requests. 

#### Issues

Related to #  
Closes #93


src/api/shared-helpers/search-issues-cache.ts
```diff
@@ -0,0 +1,47 @@
...
+	return searchIssuesCache[ serializedParams ];
+}
+
+export function setSearchIssuesCache(
+	{ search, options }: SearchParams,
+	response: SearchIssueApiResponse
+): void {
+	const serializedParams = serializeParams( { search, options } );
+	if ( cacheKeyQueue.length >= MAX_CACHE_SIZE ) {
+		const oldestCacheKey = cacheKeyQueue.shift();
```
karenroldan, line 31: **suggestion (non-blocking):**  If the `oldestCacheKey` is undefined, the cache is not cleared even if it's full. It could potentially be a source of a bug in the future! What do you think about adding some error handling here?

src/duplicate-results/sub-components/use-set-height-after-request.ts
```diff
@@ -0,0 +1,28 @@
...
+	const resultsContainerContentRef = useRef< HTMLDivElement >( null );
+	const [ resultsContainerContentHeightPx, setResultsContainerContentHeightPx ] = useState<
+		number | undefined
+	>( undefined );
+
+	useEffect( () => {
+		if (
+			resultsRequestStatus === 'fulfilled' ||
+			resultsRequestStatus === 'error' ||
+			searchTerm === ''
```
karenroldan: **suggestion (non-blocking):** If the `resultsRequestsStatus` is never `fulfilled` or `pending` when the `searchTerm` is empty, we can delete this condition.

karenroldan (approved):
Your changes are simply amazing 👏🏻  duplicate searching is soooo much smoother and more efficient. I like how the search only triggers when there's an actual change in the parameters. 

I tested switching between the `Duplicate Search` page and `Reporting Flow`, and my previous searches are still intact! This means users can easily navigate between different pages without losing their search progress. 

I've added a couple of non-blocking suggestions. Feel free to consider them or not. Otherwise, the changes look good to me! 🎉

src/api/shared-helpers/search-issues-cache.ts
dpasque: Hmmm, I'm pretty sure because of the `cacheKeyQueue.length >= MAX_CACHE_SIZE`, `oldestCacheKey` can't really ever be undefined! `shift()` only returns undefined if the array is empty.

The `if (oldestCacheKey)` check is mostly to keep the TypeScript transpiler happy, because it doesn't like the `string | undefined` type as a key indexer!

So I'm not entirely sure what extra error handling we should add here? Did you have something specific in mind?

src/duplicate-results/sub-components/use-set-height-after-request.ts
dpasque: Oh right! Good catch! Specifically, we shouldn't need the `searchTerm` in this check -- because in the case when the search term is `empty`, the request will always be what it was last (which isn't pending).  So we should be able to just check for the request status, and then include both in the dependency array!

src/api/shared-helpers/search-issues-cache.ts
dpasque: (Also, I'm going to merge this for now to do the deploy, but let me know when you end up getting a chance, and I can add this in a following fix PR! 😄 )

PR #99, opened 2023-05-16
Add ability to resize feature tree vertically

dpasque:

This make the feature tree vertically resizable! This should help users who want to do manual exploration of the tree.

Along with that, there were a couple of tiny css tweaks:
- Making the starting height 330px -- with our fixed button heights, this tends to cut an overflow button in half, which is nice for showing the fact that the list is scrollable
- I pulled the margin between the scrollable panel and the bottom panel back a bit. It felt weird to have the resize icon "floating" a fell `rem` above the bottom panel

#### Testing Instructions

Play around with it!

#### Issues

Related to #  
Closes #

### DIFF
--- a/src/feature-selector-form/feature-selector-form.module.css
+++ b/src/feature-selector-form/feature-selector-form.module.css
@@ -9,13 +9,22 @@
 	margin-bottom: 1rem;
 }
 
+.treeFieldset {
+	border: none;
+	padding: 0;
+	margin: 0;
+}
+
 .treeWrapper {
-	height: 320px;
-	overflow-y: scroll;
 	/* This helps preserve the focus outline */
 	padding: 1px;
 	margin: 0;
 	border: none;
+	resize: vertical;
+	/* This helps cut rows in half to show overflow */
+	height: 330px;
+	min-height: 80px;
+	overflow-y: scroll;
 }
 
 .firstLevel {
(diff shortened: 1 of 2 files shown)


karenroldan (approved):
LGTM! 🚀 I like this change! 

I tested it on both mobile and desktop, and resizing works well. Navigating through each tree isn't a problem. I also used VoiceOver, and it reads the tree correctly.

No issues with axe DevTools too. 👍🏻

john-legg:
Woo! This is awesome, Dan. I think this will help tremendously with tree exploration 👏 

**suggestion (if-minor):**
Would you be open to making the resize target the entire horizontal divider instead of just the corner icon? [Something like a handlebar functionality?](https://uxdesign.cc/handlebars-in-ui-design-4b36af67733b) I found the resize icon a little harder to discover and click -- especially on mobile. That's just my opinion though!

dpasque:
Hey @john-legg! 😄 

> Would you be open to making the resize target the entire horizontal divider instead of just the corner icon? [Something like a handlebar functionality?](https://uxdesign.cc/handlebars-in-ui-design-4b36af67733b) I found the resize icon a little harder to discover and click -- especially on mobile. That's just my opinion though!

Alas, I 100% couldn't agree more, but this is not a cheap change sadly. 😕  Right now we're just using the native css `resize`,  which is great because it's completely native and cheap (it's like 2 lines of css code! 😲 ), but it doesn't offer a lot of customizability.

To do a handlebar thing, we'd have to roll a JS/TS based resizer which adds quite a lot more complexity. 😕 I sadly don't think the tradeoff is there for this one!

PR #102, opened 2023-05-17
Code Hygiene: Make standardized button components for shared styling

### LINKED ISSUE #74: [Task - Cleanup]: Standardize all the different button types
### Details

We should DRY things up and standardize all our button types, now that we have a few that are re-used throughout the app (filled, outlined, non-outlined).

We could do this either with components or CSS classes

### Checklist

- [ ]

dpasque:

As I was told growing up in my American Midwest home: "the buttons that style together, stay together". 😆 

Our button styling had gotten away from us a bit! In our design, we have a shared set of a few different button styles. This PR standardizes those into a shared set of components that are then re-used throughout the app! Now the majority of the button code (css included) is all in one place! Yum! 😋 

Here's how the styling order goes...
1. In `index.css`, we target all `button` elements in the app, mostly resetting a bunch of settings to their initial styles. This is mostly to mitigate some CSS collisions when deployed internally at a8c. We also set some helpful starting points for all buttons in the site (even ones that aren't "action" buttons), like the font and a basic padding, etc.
2. In the `buttons.module.css`, we set the styles for our shared set of component buttons.
3. Other consuming components can still provide other classes to these button components to override some styles in unique cases. This almost always is just massaging `padding` -- we make certain buttons bigger or smaller in various places in the app!

What is left out?
- Buttons that are part of other controls, like the popover menu, the navbar, and the segmented control.
- The CTA links in the Next Steps. Since these are links and not true buttons, and because we treat the styling a little different for them, they seemed best left on their own.

#### Testing Instructions

The unit tests still pass, but that's not too surprising...

Mostly, go through the app and make sure all the buttons still look as they should! I've done a couple passes in this regard, but always appreciate the extra eyes 😄 

#### Issues

Related to #  
Closes #74

### DIFF
--- a/src/common/components/buttons.tsx
+++ b/src/common/components/buttons.tsx
@@ -11,15 +11,16 @@ export const OutlinePrimaryButton = createButton( styles.outlinePrimary );
 
 export const OutlineNeutralButton = createButton( styles.outlineNeutral );
 
-function createButton( buttonStlyeClass: string ) {
+function createButton( buttonStyleClass: string ) {
 	return forwardRef< HTMLButtonElement, ButtonHTMLAttributes< HTMLButtonElement > >(
 		function Button( { className, children, ...props }, ref ) {
+			const classes = [ styles.button, buttonStyleClass ];
+			if ( className ) {
+				classes.push( className );
+			}
+
 			return (
-				<button
-					className={ `${ styles.button } ${ buttonStlyeClass } ${ className }` }
-					ref={ ref }
-					{ ...props }
-				>
+				<button className={ classes.join( ' ' ) } ref={ ref } { ...props }>
 					{ children }
 				</button>
 			);
--- a/src/common/components/buttons.tsx
+++ b/src/common/components/buttons.tsx
@@ -7,6 +7,10 @@ export const ClearButton = createButton( styles.clear );
 
 export const TextButton = createButton( styles.text );
 
+export const OutlinePrimaryButton = createButton( styles.outlinePrimary );
+
+export const OutlineNeutralButton = createButton( styles.outlineNeutral );
+
 function createButton( buttonStlyeClass: string ) {
 	return forwardRef< HTMLButtonElement, ButtonHTMLAttributes< HTMLButtonElement > >(
 		function Button( { className, children, ...props }, ref ) {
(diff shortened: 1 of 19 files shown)


src/common/components/buttons.tsx
```diff
@@ -0,0 +1,28 @@
...
+export const OutlinePrimaryButton = createButton( styles.outlinePrimary );
+
+export const OutlineNeutralButton = createButton( styles.outlineNeutral );
+
+function createButton( buttonStlyeClass: string ) {
+	return forwardRef< HTMLButtonElement, ButtonHTMLAttributes< HTMLButtonElement > >(
+		function Button( { className, children, ...props }, ref ) {
+			return (
+				<button
+					className={ `${ styles.button } ${ buttonStlyeClass } ${ className }` }
```
karenroldan: **tiny nitpick:** Update the spelling from `buttonStlyeClass` to `buttonStyleClass`

src/common/components/buttons.tsx
karenroldan, line 1: **praise:**  LOVE this change! 👏🏻

src/common/components/buttons.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+export const TextButton = createButton( styles.text );
+
+export const OutlinePrimaryButton = createButton( styles.outlinePrimary );
+
+export const OutlineNeutralButton = createButton( styles.outlineNeutral );
+
+function createButton( buttonStlyeClass: string ) {
+	return forwardRef< HTMLButtonElement, ButtonHTMLAttributes< HTMLButtonElement > >(
+		function Button( { className, children, ...props }, ref ) {
```
karenroldan, line 16: **nitpick (non-blocking):** We should have a default value for the `className` so we don't get `undefined` in case we don't provide it as a prop. 

For example, in the selected feature, the class name for the `Clear` button reads as `class="buttons_button__Yk0Vz buttons_text__C85wU undefined"`

karenroldan (approved):
LGTM! 🚀 

The buttons still look as they should and I've compared it with the Bugomattic in MC 🥳  Checked for both desktop and mobile!

src/common/components/buttons.tsx
dpasque: Oops! Nice catch hahaha

src/common/components/buttons.tsx
dpasque: Oh shoot, nice catch, for some reason I thought that it would default to an empty string -- I'll get that added!

PR #105, opened 2023-05-24
Fix SegmentedControl Keyboard Nav & DRY Up the AppNavbar

### LINKED ISSUE #81: [Bug]: Make the SegmentedControl match listbox WAI ARIA keyboard standards
### Quick summary

Right now, the `SegmentedControl` component has pretty good a11y, but it doesn't match the WAI ARIA standards for `listbox`es, namely when it comes to keyboard interaction and navigation:

https://www.w3.org/WAI/ARIA/apg/patterns/listbox/

We should make it match those standards!

### Steps to reproduce

NA

### Browser

_No response_

### Other notes

_No response_

dpasque:

Previously, our `SegmentedControl` component didn't meet the WAI ARIA keyboard specs for horizontal `listbox`es: https://www.w3.org/WAI/ARIA/apg/patterns/listbox/

Now it does!

Also, I used this chance to modularize and DRY up the focus management for our `AppNavbar` and the `SegmentedControl`, because they're effectively the same!

The `AppNavbar` also now has a much more iterative structure that should make it easy to extend in the future! 🎉 

#### Testing Instructions

- [ ] Unit tests still pass
- [ ] Try tabbing into the various segmented controls and the top app navbar, and play around with the keyboard arrows!

#### Issues

Related to #  
Closes #81


src/app-navbar/app-navbar.tsx
```diff
@@ -8,113 +8,118 @@ import { ReportIssueDropdownMenu } from '../common/components/report-issue-dropd
...
+
+	const menuItems: MenuItemDetails[] = [
+		{
+			page: 'duplicate-search',
+			component: <SimpleMenuItem page="duplicate-search" label="Duplicate Search" />,
+		},
+		{
+			page: 'report-issue',
+			component:
+				issueType === 'unset' ? (
```
karenroldan: **suggestion (non-blocking):** Consider moving the `issueType` check inside the `DropdownReportIssueMenuItem`. 

What do you think about encapsulating the display logic based on `issueType` in the `DropdownReportIssueMenuItem`? The `AppNavbar` will then only need to know which items to display in the menu, not how these items should display themselves.

karenroldan (approved):
LGTM! 🚀

src/app-navbar/app-navbar.tsx
dpasque: Oh heck yeah, that reads wayyyyy nicer 😍  Great suggestion, thank you! 😄 Got it updated 👍

PR #107, opened 2023-05-25
Update existing track events

karenroldan:

Updates the existing events based on pciE2j-27d-p2

1. `type_save` - records the issue type set from duplicate -> reporting flow transition. NOTE: I'll update the name used in this flow with `navbar_report_issue_start` in adding new track events PR. 
2. `page_view` - records the query string as a property
3. `start_over_click` - records the target page as a property 

#### Testing Instructions

- [ ] Tests are passing 
- [ ] Make sure the events are recorded as expected. You can also find it in the console - `[Event recorded]:`

#### Issues

Related to #  
Closes #

### DIFF
--- a/src/common/components/report-issue-dropdown-menu.tsx
+++ b/src/common/components/report-issue-dropdown-menu.tsx
@@ -9,7 +9,6 @@ import { setIssueType } from '../../issue-details/issue-details-slice';
 import { setActiveReportingStep } from '../../reporting-flow-page/active-reporting-step-slice';
 import { setActivePage } from '../../active-page/active-page-slice';
 import { updateHistoryWithState } from '../../url-history/actions';
-import { useMonitoring } from '../../monitoring/monitoring-provider';
 
 import styles from './report-issue-dropdown-menu.module.css';
 import { selectNextReportingStep } from '../../combined-selectors/next-reporting-step';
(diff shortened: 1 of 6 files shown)


src/common/components/report-issue-dropdown-menu.tsx
```diff
@@ -52,6 +54,9 @@ export const ReportIssueDropdownMenu = forwardRef< HTMLElement, Props >(
 				// In the future, if we re-use this dropdown apart from navigation from duplicate searching,
 				// we may have defer more of these actions to calling components.
 				dispatch( setIssueType( issueType ) );
+
+				monitoringClient.analytics.recordEvent( 'type_save', { issueType: issueType } );
```
dpasque: **Update:** Ahh, sorry, I probably wasn't super clear in the planning post... But I think we should reserve `type_save` for just capturing when the user saves the type by click the "continue" button in the Reporting flow. That name is now a little generic, but I want to avoid abandoning too many track events if possible (since we've already been recording this one for that purpose), even if the names aren't quite as descriptive as they could be.

Would you be okay just leaving this out until you add the `navbar_report_issue_start` event in  a future PR?

dpasque (approved):
Awesome, thank you so much Karen! 🎉 😄 
I left one comment clarifying the `type_save` event, once that's taken care of, you should be good-to-go! 🚀

src/common/components/report-issue-dropdown-menu.tsx
karenroldan: No worries! The updated ones should be in #108! I've reverted the changes for the `type_save`. Thank you!

PR #108, opened 2023-05-25
Add track events

karenroldan:

Add track events based on pciE2j-27d-p2

1. `issue_search`
2. `status_filter_select`
3. `repo_filter_select`
4. `sort_select`
5. `issue_link_click`
6. `navbar_report_issue_start`
7. `banner_report_issue_start`
8. `navbar_item_click`

#### Testing Instructions

- [ ] Tests are passing
- [ ] Make sure the events are recorded as expected. You can also find it in the console - [Event recorded]:

#### Issues

Related to #  
Closes #

### DIFF
--- a/src/app-navbar/app-navbar.tsx
+++ b/src/app-navbar/app-navbar.tsx
@@ -120,7 +120,7 @@ const ReportIssueMenuItem = forwardRef<
 	}
 
 	return (
-		<ReportIssueDropdownMenu ref={ forwardedRef }>
+		<ReportIssueDropdownMenu ref={ forwardedRef } location="navbar">
 			<button
 				role="menuitem"
 				aria-current={ currentActivePage === 'report-issue' ? 'page' : undefined }
--- a/src/app-navbar/app-navbar.tsx
+++ b/src/app-navbar/app-navbar.tsx
@@ -8,14 +8,11 @@ import { ReportIssueDropdownMenu } from '../common/components/report-issue-dropd
 import { ReactComponent as PlusIcon } from '../common/svgs/plus.svg';
 import { ReactComponent as DownChevronIcon } from '../common/svgs/chevron-down.svg';
 import { selectIssueType } from '../issue-details/issue-details-slice';
-import { useMonitoring } from '../monitoring/monitoring-provider';
 import { useListboxFocusManager } from '../common/components';
 import { ReactElement } from 'react';
+import { useMonitoring } from '../monitoring/monitoring-provider';
 
 export function AppNavbar() {
-	const dispatch = useAppDispatch();
-	const monitoringClient = useMonitoring();
-	const issueType = useAppSelector( selectIssueType );
 	const currentActivePage = useAppSelector( selectActivePage );
 
 	interface MenuItemDetails {
@@ -84,13 +81,14 @@ const SimpleMenuItem = forwardRef< HTMLButtonElement, SimpleMenuItemProps >(
 	function SimpleMenuItem( { page, label, tabIndex }, forwardedRef ) {
 		const dispatch = useAppDispatch();
 		const currentActivePage = useAppSelector( selectActivePage );
+		const monitoringClient = useMonitoring();
 
-	const handleMenuItemClick = ( page: ActivePage ) => () => {
-		dispatch( setActivePage( page ) );
-		dispatch( updateHistoryWithState() );
+		const handleClick = ( page: ActivePage ) => () => {
+			dispatch( setActivePage( page ) );
+			dispatch( updateHistoryWithState() );
 
-		monitoringClient.analytics.recordEvent( 'navbar_item_click', { page } );
-	};
+			monitoringClient.analytics.recordEvent( 'navbar_item_click', { page } );
+		};
 
 		return (
 			<button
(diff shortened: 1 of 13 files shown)


src/app-navbar/app-navbar.tsx
```diff
@@ -81,10 +85,12 @@ const SimpleMenuItem = forwardRef< HTMLButtonElement, SimpleMenuItemProps >(
...
 
-		const handleClick = ( page: ActivePage ) => () => {
-			dispatch( setActivePage( page ) );
-			dispatch( updateHistoryWithState() );
-		};
+	const handleMenuItemClick = ( page: ActivePage ) => () => {
+		dispatch( setActivePage( page ) );
+		dispatch( updateHistoryWithState() );
+
+		monitoringClient.analytics.recordEvent( 'navbar_item_click', { page } );
```
john-legg: **issue (blocking):**
I'm not sure if you're seeing it on your end, but I actually had a couple of compile errors. One was on line `92`:

```
TS2304: Cannot find name 'monitoringClient'.
    90 | 		dispatch( updateHistoryWithState() );  
    91 |
  > 92 | 		monitoringClient.analytics.recordEvent( 'navbar_item_click', { page } );
       | 		^^^^^^^^^^^^^^^^
    93 | 	};
    94 |  
    95 | 		return (
```

I'm not sure if something happened the recent force push, but I just added `const monitoringClient = useMonitoring();` under `SimpleMenuItem` to get it working.

I also got this error on line `100`:

```
TS2304: Cannot find name 'handleClick'.
     98 | 				role="menuitem"   99 | 				aria-current={ currentActivePage === page ? 'page' : undefined }
  > 100 | 				onClick={ handleClick( page ) }
        | 				          ^^^^^^^^^^^  101 | 				tabIndex={ tabIndex }
    102 | 				className={ styles.menuItem }
    103 | 			>
```
I just changed it to `handleMenuItemClick` to get it working.

src/duplicate-search/__tests__/duplicate-search-controls.test.tsx
```diff
@@ -513,5 +570,23 @@ describe( '[DuplicateSearchControls]', () => {
...
 		} );
+
+		test( 'Selecting a sort option records event', async () => {
+			const { monitoringClient, user } = setup( <DuplicateSearchControls /> );
+
+			await user.click( screen.getByRole( 'combobox', { name: 'Sort results by…' } ) );
+			await user.click( screen.getByRole( 'option', { name: 'Date added' } ) );
+
+			expect( monitoringClient.analytics.recordEvent ).toHaveBeenCalledWith( 'sort_select', {
+				sortOption: 'date-created',
```
john-legg, line 581: **nitpick (non-blocking):**
I noticed that the event name is `date-created`, whereas the filter in the UI is `Date added`. This made me realize that we've been a bit inconsistent with our naming of this field because even the duplicate search result row uses `Opened 4 days ago` lol.

I feel like we should probably pick one and keep it the same throughout duplicate search 😆 For example, the filter and event name could be `date created` and the results row could display `Created 4 days ago`. 

Karen and @dpasque, what are your thoughts on this?

john-legg (commented):
Nice work, Karen! I had a few compile errors that I pointed out below. I fixed them locally, so I could continue testing your PR, but please take a look when you have a chance!

Once I got Bugomattic working, I checked all of the events and they fired as expected. I just had one comment about naming, but that's it.

src/app-navbar/app-navbar.tsx
karenroldan: > I had a few compile errors that I pointed out below

AAAAAH. My bad! I had two VS Codes (one is the Insiders one since I'm playing with the Copilot chat) open and didn't see the error after "fixing" the merge conflicts! It should be fixed now. 

Thank you for the heads up! 🙇🏻‍♀️

src/duplicate-search/__tests__/duplicate-search-controls.test.tsx
karenroldan: > I feel like we should probably pick one and keep it the same throughout duplicate search 😆 For example, the filter and event name could be date created and the results row could display Created 4 days ago.

That's a good point! Since `date created` is used mostly everywhere, we can update the text in the dropdown from `Date added` to `Date created`.

karenroldan: Created a separate PR for this change: https://github.com/Automattic/bugomattic/pull/110

john-legg:
This looks great, Karen! The code compiled properly and all of the track events work as expected 😄 

I did want to follow up on one note I made in the [previous comment](https://github.com/Automattic/bugomattic/pull/108#discussion_r1207346059)

> This made me realize that we've been a bit inconsistent with our naming of this field because even the duplicate search result row uses Opened 4 days ago lol.

For the duplicate search result rows, would it be possible to change the word `Opened` to `Created` to better match the renamed `Date Created` sorting option? This isn't urgent, so it can be in a separate PR if the word change sounds reasonable to you!

john-legg (approved):

karenroldan:
> For the duplicate search result rows, would it be possible to change the word Opened to Created to better match the renamed Date Created sorting option?

Ofc, I missed the previous text! 🤦🏻‍♀️ it should be fixed in #113. Thanks again for checking.

PR #111, opened 2023-05-30
Add support for JIRA tickets

### LINKED ISSUE #85: [Task - Reporting Flow]: Add support for JIRA tickets
### Details

Tumblr uses them - we should add a task type for this!  👍  Probably need to reach out to Tumblr to get the details of how they do that issue tracking

### Checklist

- [ ]

karenroldan:

Adds a task type for JIRA. 

Screenshot 2023-06-01 at 11 52 00

#### Testing Instructions

* Ensure that the tests are passing 

#### Issues

Related to #  
Closes #85


src/common/lib/links.ts
```diff
@@ -67,3 +68,11 @@ export function createNewGithubIssueHref( link: NewGitHubIssueLink, issueTitle?:
 
 	return url.href;
 }
+
+export function createNewJiraIssueHref( link: NewJiraIssueLink ) {
+	const url = new URL( 'https://jira.tumblr.net/secure/CreateIssueDetails!init.jspa' );
+	// If the project ID or issue type is invalid, the link will just show an error saying as much.
+	url.searchParams.append( 'pid', link.project.id.toString() );
+	url.searchParams.append( 'issuetype', link.issueType.id.toString() );
+	return url.href;
```
dpasque: **Suggestion (blocking):** Hmm, I'm wondering if we want to avoid hardcoding the URL here 🤔 I believe other teams in the past other than Tumblr have used Jira, and I could see us maybe needing to broaden our support from just Tumblr here. My guess is that all new issues use the same path (`/secure/CreateIssueDetails!init.jspa`), so maybe we could add the host name of the JIRA instance (in this case, `https://jira.tumblr.net`) to the configuration type?

What do you think?

src/static-data/reporting-config/types.ts
```diff
@@ -114,7 +114,23 @@ export interface NewGitHubIssueLink {
...
+
+interface JiraIssueType {
+	name: string;
+	id: number;
+}
+
+interface JiraProject {
+	name: string;
+	id: number;
+}
```
dpasque: **Suggestion: (non-blocking)** What do you think of simplifying this data structure down to just use the ids?

So it would look like:
``` typescript
export interface NewJiraIssueLink {
	type: 'jira';
	projectId: number;
	issueTypeId: number;
}
```

Definitely don't feel strongly here, and mostly want to open up for discussion!

I really like the idea here, but I wonder if it's the right tradeoff for the reporting config. 🤔  I've been trying to keep this data structure as simple as possible to minimize the number of fields people have to fill out to populate the field.

Technically, the only things that really matter are the IDs, as they are used in the URL creation. All the other stuff is just used to add some niceness to the default text.

But... Since that is all manually entered, the name could actually get out of sync with its actual value on the JIRA server. (Like if project with id `123` was called `Foo Project`, but later is updated to `Bar Project`, what we would show in Bugomattic wouldn't match where the URL takes folks).

So it's a small step backwards in terms of user experience, but I think the simplicity might be nice in the long run.

Thoughts?

src/common/svgs/jira.svg
```diff
@@ -0,0 +1 @@
+<svg height="2500" preserveAspectRatio="xMidYMid" width="2500" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 -30.632388516510233 255.324 285.95638851651023"><linearGradient id="a"><stop offset=".18" stop-color="#0052cc"/><stop offset="1" stop-color="#2684ff"/></linearGradient><linearGradient id="b" x1="98.031%" x2="58.888%" xlink:href="#a" y1=".161%" y2="40.766%"/><linearGradient id="c" x1="100.665%" x2="55.402%" xlink:href="#a" y1=".455%" y2="44.727%"/><path d="M244.658 0H121.707a55.502 55.502 0 0 0 55.502 55.502h22.649V77.37c.02 30.625 24.841 55.447 55.466 55.467V10.666C255.324 4.777 250.55 0 244.658 0z" fill="#2684ff"/><path d="M183.822 61.262H60.872c.019 30.625 24.84 55.447 55.466 55.467h22.649v21.938c.039 30.625 24.877 55.43 55.502 55.43V71.93c0-5.891-4.776-10.667-10.667-10.667z" fill="url(#b)"/><path d="M122.951 122.489H0c0 30.653 24.85 55.502 55.502 55.502h22.72v21.867c.02 30.597 24.798 55.408 55.396 55.466V133.156c0-5.891-4.776-10.667-10.667-10.667z" fill="url(#c)"/></svg>
```
dpasque: **Question:** Out of curiosity, where did you get this SVG? If possible, we should use the official one from their press kit -- there should be a blue version of the icon in the zip here: https://assets.ctfassets.net/8j5aqoy0ts8s/1y9Zan62VaterUggPDIDAx/7537cd8851766dfce84d0785dc8200d7/Jira.zip

I just ask because I've been trying to be cautious with official org logos and to use the official press kit when possible 😄 , and this one seemed a little different from that one in the press kit! (And please ignore if this is from there and I'm making the comparison incorrectly! 🙇 )

dpasque (commented):
Hurray, thank you so much for doing this! 🎉  I had a few comments and questions, let me know what you think!

src/common/svgs/jira.svg
karenroldan: > I just ask because I've been trying to be cautious with official org logos and to use the official press kit when possible

That's a valid point! I got it from the first searches on Google. I've replaced the SVG to use the official one 🙇🏻‍♀️ Fix can be found here: [3beaeed](https://github.com/Automattic/bugomattic/pull/111/commits/3beaeed3e0978ae9eda3a66cd8c38acef13d4f47)

src/static-data/reporting-config/types.ts
karenroldan: > Technically, the only things that really matter are the IDs, as they are used in the URL creation. All the other stuff is just used to add some niceness to the default text. 

Yes, that's right! I was contemplating between requiring the name for readability vs id only. 

> So it's a small step backwards in terms of user experience, but I think the simplicity might be nice in the long run. Thoughts?

I totally agree with you and I'm okay with simplifying the data structure. It makes sense that it will be an overhead once the project name is updated and we don't want to have that maintenance on the reporting config. 

I appreciate the feedback! I'll get that fixed 👍🏻

src/common/lib/links.ts
karenroldan: > so maybe we could add the host name of the JIRA instance (in this case, https://jira.tumblr.net) to the configuration type? 

I see your point! I've updated the structure and added the host name for Jira. 👍🏻

src/next-steps/sub-components/task.tsx
```diff
@@ -160,7 +160,7 @@ function getDefaultTitleForLink( link: TaskLink ): string {
 		case 'github':
 			return `Open an issue in the ${ link.repository } repo`;
 		case `jira`:
-			return `Open an issue in the ${ link.project.name } project`;
+			return 'Open a new issue';
```
karenroldan, line 163: **context:** p1685613224978629-slack-CQD1HH4MA

john-legg (approved):
Thanks for adding this feature, Karen! The code looks good and my testing didn't find any issues. 🚀

PR #110, opened 2023-05-29
Update label from 'Date added' to 'Date created'

karenroldan:
#### What Does This PR Add/Change?

* Updates the label from `Date added` to `Date created` for consistency. 
* Context: https://github.com/Automattic/bugomattic/pull/108#discussion_r1207346059
Screenshot 2023-05-29 at 08 38 41

#### Testing Instructions

1. Click on the sort option 
2. Verify that the label displayed is `Date created` 
3. Check the query string also and make sure that the URL still contains `&duplicateSearch.sort=date-created`

#### Issues

Related to #  
Closes #

PR #113, opened 2023-06-02
Use 'Created'  in issue result

karenroldan:

Related to #110 and follow-up from #108

With the changes on the label from 'Date added' to 'Date created' in #110, we also need to update the text in the issue results! 

Now the copy should be consistent across the app! The text for the date created display should now use `Created`

Screenshot 2023-06-02 at 09 37 16

#### Testing Instructions

1. Search for any issue 
2. Verify that the result says `Created x days ago` or `Created x months ago`

#### Issues

Related to #  
Closes #


john-legg (approved):
Looks great, Karen! Thanks for making the change 🚀

PR #115, opened 2023-06-14
Tune up description searching match criteria

dpasque:

After playing around with the real reporting data, I felt like our current description match logic (just any single token needs to match) was a _bit_ too generous -- you can get a **lot** of hits sometimes, and they aren't all meaningful!

So, to help with that, I made two changes...
1. I migrated our stop word filtering to the library `[stopword](https://www.npmjs.com/package/stopword)` -- it's a bit more comprehensive than our current list, and is now one less thing for us to maintain! 🎉 
2. We now have a stricter threshold to clear. The ratio of (**the total number of hits in the description**) / (**the number of meaningful tokens in the search term**) must be >= **0.6 (60%).**

Why did that I take that approach? Several notes...

- It's common practice to use a ratio threshold in these kinds of searches. This allows the amount to scale with the length of the search.
- I specifically wanted the case where if you used two search terms, you needed two search hits, but I didn't want to overshoot too far, which pushed me to 0.6.
- I wanted to use total hits in the numerator in case a word comes up a lot in the description
- I wanted to use meaningful words in the denominator so that stop words don't artificially inflate the requirements and make things too strict

#### Testing Instructions

Unit tests should still pass!

Best way to test is to use a semi-realistic reporting config locally `REACT_APP_REPORTING_CONFIG_NAME='realistic-config' yarn start`, and just play around and trust your gut! How does it feel? Do we need to pull back or push further?

#### Issues

Related to #  
Closes #

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,6 +71,7 @@
 		"react-redux": "^8.0.4",
 		"react-scripts": "5.0.1",
 		"react-tooltip": "^5.8.3",
+		"stopword": "^2.0.8",
 		"url-join": "^5.0.0",
 		"web-vitals": "^2.1.0"
 	},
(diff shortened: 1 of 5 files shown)


karenroldan (approved):
Love this change! 😍 you now get the best results based on what you typed in! I think the threshold is a good start. 

I like how the displayed results are more connected to what was searched for. The current match level gives more accurate and relevant results. Thank you for improving it! 🙇🏻‍♀️

PR #123, opened 2023-06-21
Rename duplicate searching to issue searching everywhere

### LINKED ISSUE #116: [Task]: Switch "Duplicate Search" --> "Search for Issues"
### Details

Should include:
- Name of activity
- Header of page
- State value in URL

Maybe should do in all the file names too.

p1681925679608289-slack-CQD1HH4MA

### Checklist

- [ ]

karenroldan:
Duplicate searching is a bit narrow of a term, and it would be better to reference just "issue searching" or "searching for existing issues".

Since we're still pre-launch, let's just knock this all out at once and rename everything too, so that everything reads consistently and isn't confusing to future developers!

This changes include:
- Renaming the header, the page value in the url state, and all the copy/text re: issue searching.
- Renaming all the files, and folders to remove "duplicate"
- Rename all the component names, other function names, and the redux state nodes to remove "duplicate" from it.

#### Testing Instructions

- Make sure there are no compilation issues 
- All tests are passing
- Manual regression test passes

**Verify the following:**

1. The page name is `Search for Issues`
2. Active page in the URL says `search-for-issues`
3. Track event page name is `search-for-issues`

#### Issues

Related to #  
Closes #116


dpasque (approved):
Made a few more tweaks and went over everything:
- tests are passing
- app functionally looks good
- voice over reads well

Issue #134, opened 2023-06-23
[Bug]: Diagnose and fix performance of the feature search


dpasque:
### Quick summary

The feature tree is starting to drag a bit with the current size of the reporting config.

We should figure out where the slowdowns are happening and speed it up!

### Steps to reproduce

Interact with the feature tree with searches. You'll eventually notice it drag.

Also, the chrome console is letting us know there are some long event loop runs happening 

### Other notes

I'll profile to figure out what the problem is: whether it's the substring search or the rendering of too many different components in the tree (if the search is too permissive and returns too many results).

I'm pretty confident it's the react rendering of too many components that is the issue.

dpasque:
After some more manual investigation, it seems like the bad performance occurs when...
1. You have a lot of visible items in the tree
2. You try a new search.

So for example, expand everything manually, then do a new search. You'll notice the hang!

dpasque:
This has nothing to do with the search -- that runs literally in 1 ms haha

It's the react commit phase (where it's updating the DOM).

There's just so many components here getting rendered. And the problem is that the `react-tooltip` library we are using doesn't seem to scale well. If you take it out, the performance improves a lot, but still isn't totally there. 

So, some next steps...
1. See if we can reduce the number of components being rendered/managed in the DOM. Maybe we can use virtualization here.
2. See if we need to find a more performance tooltip library if the above doesn't seem to fix it.

dpasque:
FYI to @karenroldan as this is probably one of the higher priority things to try to fix during this CfT 😅

dpasque:
Okay, lots of updates!

In the ideal world, when you have cases like this where the React commit phase is too long, virtualization is a good approach. However, virtualization is hard here -- a lot of those libraries expect flattened lists, and with how our expandable/collapsible/searchable tree works, flattening it into something to be virtualized while preserving that behavior is just tough!

We still might be able to do virtualization in the future, but it would require a complete refactor of how we build this tree.

So, to start, I think we should see how far we can get without that by optimizing how the tree responds to searches.

First step is PR #135 -- it turns out that part of what is so slow is doing the reconciliation pass when updating the DOM. It's actually way faster to just toss out everything and rebuild the tree from scratch on each new search term. That already matches with the user experience.

Next, I think we need to minimize:
1. The number of results that are returned for any given search, and...
2. The number of components involved in building out that tree

[...]

PR #136, opened 2023-06-26
Tree Performance Round 2: Add character limit for feature search

karenroldan:

This is part of optimizing the performance of the feature search. 

When searching for a feature, **there is now a minimum of 3 characters to trigger the search.** The behavior is similar when searching for issues:

- you will need to press enter to bypass debouncing
- if the search term is empty, it will not trigger a search

#### Testing Instructions

- Make sure all tests passed 
- A search term with 3 characters should trigger the search. Check that the text highlight is still working as expected 

#### Issues

Related to #134 
Closes #

### DIFF
--- a/src/feature-selector-form/__tests__/feature-selector-search.test.tsx
+++ b/src/feature-selector-form/__tests__/feature-selector-search.test.tsx
@@ -372,5 +372,32 @@ describe( '[FeatureSelector -- Tree interaction]', () => {
 				screen.getByText( 'No results found. Try a different search or explore manually below.' )
 			).toBeInTheDocument();
 		} );
+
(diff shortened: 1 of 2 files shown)


dpasque:
@karenroldan This is fantastic!! This makes me realize I was probably way over-thinking this too haha... Having the debounce limit avoids the more silly cases while still allowing people to override it!

I tweaked the debounce time back to 300ms btw -- from research online, this seems more in the range of what it should be for a quick client-side search. I wanted to avoid too long too because we don't have a loading indicator!

I'm going to merge this FYI! 

Thank you again -- this feels really good and makes a big difference! 🚀 

FYI to @john-legg that this is how we're helping getting around the silly one-letter searches

PR #148, opened 2023-07-14
Make all localstorage caches error safe

### LINKED ISSUE #147: Error when loading the reporting config
### Quick summary

Context: pb6Nl-gEx-p2#comment-103208

```
QuotaExceededError: Failed to execute 'setItem' on 'Storage': Setting the value of 'cachedReportingConfigData' exceeded the quota.
``` 

### Steps to reproduce

1. Store more than the allocated space in the local storage of the browser
2. Load Bugomattic 
3. A message "Uh oh! Something went strong :(" will be displayed.

### Browser

_No response_

### Other notes

_No response_

dpasque:

All our localstorage cache sets are now error safe, meaning a failure to cache will never disrupt site usage.

Now, we just leave a helpful console message that the user might be able to speed up the site by making some room in localstorage

#### Testing Instructions

This has to be tested on an internal deployment to use the real API -- I've taken care of that! I deployed on my sandbox and inflated local storage so there was no room, and verified all 4 keys don't throw errors and leave the console message.

Also, all unit tests are passing!

#### Issues

Related to #  
Closes #147

### DIFF
--- a/src/api/production-api-client.ts
+++ b/src/api/production-api-client.ts
@@ -37,8 +37,8 @@ class ProductionApiClient implements ApiClient, LoggerApiClient {
 			const data = await response.json();
 
 			const cacheExpiryTime = Date.now() + 3 * 24 * 60 * 60 * 1000; // 3 days in milliseconds
-			localStorage.setItem( 'cachedReportingConfigData', JSON.stringify( data ) );
-			localStorage.setItem( 'cacheExpiry', cacheExpiryTime.toString() );
+			safelySetLocalstorageCache( 'cachedReportingConfigData', JSON.stringify( data ) );
+			safelySetLocalstorageCache( 'cacheExpiry', cacheExpiryTime.toString() );
 
 			return data;
 		} else {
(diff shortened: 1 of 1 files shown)


src/api/production-api-client.ts
```diff
@@ -164,3 +164,18 @@ export function createProductionApiClient(): ApiClient & LoggerApiClient {
...
+function safelySetLocalstorageCache( key: string, value: string ) {
+	try {
+		localStorage.setItem( key, value );
+	} catch ( err ) {
+		// Caching should never bubble up errors -- it's just a performance optimization!
+		// The most likely thing we'd hit when caching is a QuotaExceededError.
+		// So let's just leave a gentle console message about the user being able to add a perfomrance boost.
+		console.info(
+			`Unable to set cache value for ${ key }. ` +
+				`This likely means there is too much in localstorage for this domain. ` +
```
karenroldan: **suggestion (non-blocking):** Replace 'domain' with 'website'. I know we love domains lately (lol), but I think using ´website´ here will be more user-friendly!

> This likely means there is too much in localstorage for this domain

karenroldan (approved):
Yay! Thank you so much for looking into this! The changes look good to me. 🚀

src/api/production-api-client.ts
dpasque: Bahaha, you're totally right! 😆 I'll get that updated.

PR #206, opened 2024-10-04
Default page: change default page to bug reporting page

jeherve:

Instead of landing in Bugomattic via the Issue search screen, let's go directly to the issue reporting screen. 

This was requested by @inaikem

#### Testing Instructions

When first loading the app, it should load the issue reporting screen right off the bat. You should still be able to switch to the issue search screen.


jeherve:
Deployed to WordPress.com in r305892-wpcom.

